PR #8, opened 2023-02-07
feat(db schema): [OSL-142, OSL-143] Add List and ListItem models

nina-py:
## Goal

Set up a Prisma schema for the success of this project! Only slightly exaggerating here. The more we get right at the start, the easier it will be down the track. I have refrained from running the first migration to actually create the tables while the schema is still in flux.

- Added the List model with all the required fields + proposed moderation-related fields.

- Added the ListItem model with all the required fields.

- Added two enums for `Status` and `ModerationStatus`.

- Formatted the schema and validated it with Prisma.

## Tickets

- https://getpocket.atlassian.net/browse/OSL-142
- https://getpocket.atlassian.net/browse/OSL-143


katerinachinnappan:
I thin it will be nice to add `prisma generate` and `prisma migrate deploy` to `package.json` to deploy the models after they are finalized in this PR. Can either add it to this PR or make another small one and run the migration

jpetto (approved):
put it on the ship!

PR #9, opened 2023-02-07
chore(node): [OSL-173] Update Node version to v.18 + upgrade key packages

nina-py:
## Goal

- Updated Node, @types/node, Prisma and typescript packages manually for both the infrastructure code and the main repository. The rest will be taken care of by the Renovate bot with time.

## Tickets

https://getpocket.atlassian.net/browse/OSL-173


katerinachinnappan (commented):
Should we also update CircleCI jobs to use node 18? Currently some jobs run on node 16:
`image: circleci/node:16@sha256:1803e9ed7deec9456ad2609124b7333d40b2eec0cf34998ae766cbf90c9a3625`

nina-py:
@katerinachinnappan, thanks for spotting that! Updated + updated the app image on Docker to also use Node 18. All checks are passing now that I've merged `main` to add recent changes.

.circleci/config.yml
```diff
@@ -46,7 +46,7 @@ only_dev: &only_dev
 jobs:
   apollo:
     docker:
-      - image: circleci/node:16@sha256:1803e9ed7deec9456ad2609124b7333d40b2eec0cf34998ae766cbf90c9a3625
+      - image: cimg/node:18.14
```
nina-py, line 49: Renovate bot will pin it to a specific build later, for now the most pertinent change is using the new CircleCI images (`cimg` instead of `circleci`).

katerinachinnappan: @nina-py I chatted with @cmharlow and I think we want to avoid using `cimg`? So instead: `circleci/node:18-slim@sha256:0a621cdd7d66ad8976f4246ab0661e3b1dd0d397c1dd784ea01bf740bd1c2522`

nina-py: Thanks Katerina! The checks failed with the slim image, and there is no full Node 18 image under `circleci` tag in the Docker repository - they've been moving over to `cimg` for some time. I switched to using a stock standard node image and that works! 🎉

katerinachinnappan (approved):
great work!!🔥

jpetto (approved):

PR #18, opened 2023-02-08
feat(server): [OSL-150] Add context, split into admin and public servers

nina-py:
## Goal

The original goal was to update Apollo packages to v.4, but @katerinachinnappan already did this in a previous PR, which simplified the task greatly.

- Moved existing code and mock resolvers into `src/public`.

- Added an admin server with its own schema.

- Had to add the `prisma.schema` file as the addition of the DB connection in the server context meant Prisma types have to be existing for the servers to run.

- Also added placeholder code for access groups - `READONLY` is for the existing Pocket group and the `FULLACCESS` one is to be confirmed and actually created when it's needed.

## Tickets

https://getpocket.atlassian.net/browse/OSL-150


katerinachinnappan:
looks great! Can we test this in dev first to make sure the api get deployed and we can get the dummy data?

katerinachinnappan:
I pushed this branch to dev and deployment is stable. approved!

katerinachinnappan (approved):
🔥

PR #21, opened 2023-02-08
fix(schema): [OSL-185] Add `url` field to Prisma schema

nina-py:
## Goal

Tweak the schema to add URLs for each saved item.

- Added the `url` field for `ListItem` model that was missed in the first iteration of the schema.

- Added the initial Prisma migration that creates tables for the two models.

## Tickets

- https://getpocket.atlassian.net/browse/OSL-185


sri-kirsh (approved):

prisma/migrations/migration_lock.toml
```diff
@@ -0,0 +1,3 @@
+# Please do not edit this file manually
+# It should be added in your version-control system (i.e. Git)
+provider = "mysql"
```
sri-kirsh, line 3: nit: eof

nina-py: It's an auto-generated file, so I think it's fine to leave it as is

prisma/migrations/20230208200307_init/migration.sql
```diff
@@ -0,0 +1,41 @@
...
+    PRIMARY KEY (`id`)
+) DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+
+-- CreateTable
+CREATE TABLE `ListItem` (
+    `id` BIGINT NOT NULL AUTO_INCREMENT,
+    `externalId` VARCHAR(255) NOT NULL,
+    `listId` BIGINT NOT NULL,
+    `itemId` INTEGER NULL,
+    `url` VARCHAR(500) NULL,
```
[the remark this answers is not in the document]
nina-py: Text is for when we're sure we cannot put an upper limit on the contents of a field.

Herraj (approved):

prisma/migrations/20230208200307_init/migration.sql
```diff
@@ -0,0 +1,41 @@
...
+    `title` VARCHAR(300) NULL,
+    `excerpt` TEXT NULL,
+    `imageUrl` TEXT NULL,
+    `authors` TEXT NULL,
```
jpetto, line 31: (sorry i missed this in the last PR) - we could do a varchar here. it'll likely be a stringified json object. i bet we could do a 1000 char limit and be totally fine.

prisma/migrations/20230208200307_init/migration.sql
```diff
@@ -0,0 +1,41 @@
...
+-- CreateTable
+CREATE TABLE `ListItem` (
+    `id` BIGINT NOT NULL AUTO_INCREMENT,
+    `externalId` VARCHAR(255) NOT NULL,
+    `listId` BIGINT NOT NULL,
+    `itemId` INTEGER NULL,
+    `url` VARCHAR(500) NULL,
+    `title` VARCHAR(300) NULL,
+    `excerpt` TEXT NULL,
+    `imageUrl` TEXT NULL,
```
jpetto, line 30: i think we could also do a varchar here. in collection api we have a 500 varchar for image urls. we could follow the same - or even up it to 1000 if we wanted to feel Very Safe.

jpetto (changes requested):
realizing i'm actually asking for changes on the previous PR - sorry!

jpetto (approved):

PR #23, opened 2023-02-08
OSL-144: adding public api schema + List & ListItem types

katerinachinnappan:
## Goal

* Added `List` and `ListItem` types to the public schema

## Todos

- [ ] deploy to dev first

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-144](https://getpocket.atlassian.net/browse/OSL-144)


schema-public.graphql
```diff
@@ -1,19 +1,36 @@
...
+	userId: String!
+	slug: String!
+	title: String!
 	description: String
-	status: String
-	listItems: [ListItem]
+	status: ListStatus!
+	moderationStatus: ModerationStatus!
+	moderatedBy: String
+	moderationReason: String
```
nina-py, line 32: We should not be exposing `moderatedBy` and `moderationReason` props on the public graph.

schema-public.graphql
```diff
@@ -1,19 +1,36 @@
...
+	sortOrder: Int!
 }
 
 type List {
-	id: String
-	title: String
-	author: String
-	slug: String
+	externalId: ID!
+	userId: String!
```
nina-py, line 25: Should we be exposing this on the public graph? I think not.

schema-public.graphql
```diff
@@ -1,19 +1,36 @@
...
+	VISIBLE
+	HIDDEN
+}
+
 type ListItem {
-	id: String
-	listId: String
-	itemId: String
+	externalId: ID!
+	listId: Int!
```
nina-py, line 13: We don't need to expose this, I believe

nina-py (commented):
Great work! We should remove a few properties that are for internal consumption only, otherwise all good! 🚀

nina-py (approved):

PR #27, opened 2023-02-10
feat(query): OSL-146 Add `shareableList` query

nina-py:
## Goal

- Updated the public schema to remove fields that shouldn't be exposed on the public graph

- Added resolvers for `shareableList` query and return types.

## Todos

- [x] Get feedback on the schema changes
- [x] Get feedback on the shape of the proposed query (with two variables) is what we need
- [x] Address review feedback
- [x] Add helper functions for `List` and `ListItems` for generating data for integration tests
- [x] Add integration tests
- [x] Swap `slug` for `externalId` in query variables (we won't be creating the slug until much later in the shareable list life cycle).

## Tickets

-  https://getpocket.atlassian.net/browse/OSL-146


schema-public.graphql
```diff
@@ -1,36 +1,50 @@
...
 
 type Query {
-	lists: [List]
+	lists: [ShareableList]
+
+  """
+  Looks up and returns a Shareable List with a given slug for a given user
+  """
+  shareableList(userId: String!, slug: String!): ShareableList
 }
```
nina-py, line 50: Does this look about right? Should we be batching up variables in some sort of `Input` type, or is this okay as is given there are only two variables?

schema-public.graphql
```diff
@@ -1,36 +1,50 @@
-enum ListStatus {
+"""
+ISOString scalar - all datetimes fields are Typescript Date objects on this server &
+returned as ISO-8601 encoded date strings (e.g. ISOString scalars) to GraphQL clients.
+See Section 5.6 of the RFC 3339 profile of the ISO 8601 standard: https://www.ietf.org/rfc/rfc3339.txt.
+"""
+scalar ISOString
```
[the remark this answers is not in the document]
jpetto: i think we should continue using this format. 👍

schema-public.graphql
```diff
@@ -1,36 +1,50 @@
...
+enum ShareableListStatus {
 	PRIVATE
 	PUBLIC
 }
 
-enum ModerationStatus {
+enum ShareableListModerationStatus {
 	VISIBLE
 	HIDDEN
 }
```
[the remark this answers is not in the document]
jpetto: great catch!

schema-public.graphql
```diff
@@ -1,36 +1,50 @@
...
 	description: String
-	status: ListStatus!
-	moderationStatus: ModerationStatus!
-	moderatedBy: String
-	moderationReason: String
-	listItems: [ListItem!]!
+	status: ShareableListStatus!
+	moderationStatus: ShareableListModerationStatus!
+  createdAt: ISOString!
+  updatedAt: ISOString!
```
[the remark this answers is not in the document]
jpetto: no worries nina - this is why we iterate 😄

schema-public.graphql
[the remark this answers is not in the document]
katerinachinnappan: Yeah I think this is a newer version so lets stick to this!

schema-public.graphql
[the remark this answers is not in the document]
katerinachinnappan: Good catch!

schema-public.graphql
katerinachinnappan: This is ok! I did the same with `Labels` in collection-api.

src/database/types.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+export type ShareableListItem = Omit<
+  ListItem,
+  | 'id'
+  | 'externalId'
+  | 'listId'
+  | 'itemId'
+  | 'sortOrder'
+  | 'createdAt'
+  | 'updatedAt'
```
jpetto, line 15: any reason we shouldn't send `sortOrder`, `createdAt`, and `updatedAt`? looks like we're retrieving `sortOrder` in the query resolver above, and created/updated might be nice metadata for clients (that they can of course omit).

schema-public.graphql
```diff
@@ -1,36 +1,50 @@
...
-	itemId: Int
-	url: String
+type ShareableListItem {
+	url: Url
 	title: String
 	excerpt: String
-	imageUrl: String
+	imageUrl: Url
 	authors: String
 	sortOrder: Int!
```
jpetto, line 29: any reason we shouldn't include `createdAt` and `updatedAt` as available fields here?

jpetto (changes requested):
i think only one bug (omitting `sortOrder` from the db type), and a suggestion to maybe include the created/updated at fields for list items.

love this quick turnaround clean up!

src/public/index.ts
```diff
@@ -23,12 +25,14 @@ for (let i = 0; i < 5; i++) {
...
 }
 
+// dummy resolver
 function lists(): any {
   return listData;
 }
-// dummy resolver
-export const resolvers = {
+
+export const index = {
```
katerinachinnappan, line 33: not blocking but we could probably leave this as `resolvers` to be more intuitive. same with the file name change

src/public/resolvers/queries/ShareableList.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @param userId
+ * @param slug
+ * @param db
+ */
+export async function getShareableList(
+  parent,
+  { userId, slug },
+  { db }
+): Promise<ShareableList> {
+  const list = await dbGetShareableList(db, userId, slug);
```
katerinachinnappan, line 16: are we getting the  `userId` from the headers?

schema-public.graphql
jpetto: i think this is okay for the use cases we can see. getting a single entity _shouldn't_ have many filter options.

src/public/resolvers/queries/ShareableList.ts
jpetto: good catch @katerinachinnappan! we likely need a context manager here to pull the header value out.

katerinachinnappan: @jpetto yep, I have that in my PR for creating a list!

src/database/types.ts
nina-py: Good point - added those

src/public/index.ts
nina-py: Good catch, this was meant to go in the `/resolvers/` folder hence the renaming to `index.ts`. The IDE updated the variable name for me and I didn't notice.

schema-public.graphql
nina-py: Added them in!

src/public/context.ts
```diff
@@ -15,17 +15,25 @@ export interface IPublicContext {
...
-  get userId(): string {
+  get userId(): IPublicContext['userId'] {
     const userId = this.config.request.headers.userid;
+
+    // We need this check for every query and mutation on the public graph
+    if (!userId) {
+      throw new AuthenticationError(
+        'You must be logged in to use this service'
+      );
+    }
```
nina-py, line 35: I have moved this from the mutation to the context as we need this check everywhere

src/public/resolvers/mutations/ShareableList.integration.ts
```diff
@@ -46,7 +43,10 @@ describe('mutations: List', () => {
   describe('createShareableList', () => {
     beforeAll(async () => {
       // Create a List
-      await createShareableListHelper(db, 'Simon Le Bon List');
+      await createShareableListHelper(db, {
+        userId: headers.userId,
+        title: 'Simon Le Bon List',
+      });
```
nina-py, line 49: I updated integration tests for the mutation as they needed changes following a refactor of the helpers.

src/test/helpers/createShareableListHelper.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Generates a Shareable List with supplied or randomised values.
+ *
+ * @param prisma
+ * @param data
+ */
+export async function createShareableListHelper(
+  prisma: PrismaClient,
+  data: ListHelperInput
+): Promise<List> {
```
nina-py, line 22: Updated this to return Prisma types rather than types that match the public schema - this way the returned object can be used in more flexible ways (e.g., for admin graph integration tests, too), and also has props we need to link list items to lists.

schema-public.graphql
```diff
@@ -48,6 +50,12 @@ input CreateShareableListInput {
 
 type Query {
 	lists: [ShareableList]
+
+  """
+  Looks up and returns a Shareable List with a given slug for a given user
+  (the user ID will be coming through with the headers)
+  """
+  shareableList(slug: String!): ShareableList
```
jpetto, line 58: for this auth-only query, we'll use `externalId` instead of `slug`. apologies that this wasn't made clear in the ticket. 😞

src/config/index.ts
```diff
@@ -15,4 +15,5 @@ export default {
     release: process.env.GIT_SHA || '',
     environment: process.env.NODE_ENV || 'development',
   },
+  slugify: { lower: true, remove: /[*+~.()'"!:@]/g },
```
jpetto, line 18: nice

src/public/context.ts
jpetto: there's _one_ place we won't need this, and that's on the `publicShareableList` query.

src/test/helpers/createShareableListItemHelper.ts
```diff
@@ -0,0 +1,39 @@
...
+  data: ListItemHelperInput
+): Promise<ListItem> {
+  const input = {
+    listId: data.list.id,
+    url: data.url ?? `${faker.internet.url()}/${faker.lorem.slug(5)}`,
+    title: data.title ?? faker.random.words(5),
+    excerpt: data.excerpt ?? faker.lorem.sentences(2),
+    imageUrl: data.imageUrl ?? faker.image.cats(),
+    authors: data.authors ?? faker.name.fullName(),
+    sortOrder: data.sortOrder ?? faker.datatype.number(),
```
jpetto, line 33: no changes needed, but just calling out that `listId`, `url`, and `sortOrder` are the only guaranteed fields (ty parser).

jpetto (changes requested):
this looks really good - i think the only required update is changing from `slug` to `externalId` as the lookup param.

schema-public.graphql
nina-py: Fixed!

jpetto (approved):
looks good!

PR #28, opened 2023-02-10
OSL-95: adding createShareableList mutation

katerinachinnappan:
## Goal

* Adding a `createShareableList` mutation
* Adding integration test
* Updating public schema and db types per @nina-py suggestion in her PR
* Updating prisma schema and making `slug` optional on `List` type (we don't provide the `slug` when creating a List)

## Todos

- [ ] deploy to dev
- [x] Fix CircleCI test_integration error

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-95](https://getpocket.atlassian.net/browse/OSL-95)


schema-public.graphql
```diff
@@ -1,36 +1,59 @@
-enum ListStatus {
+"""
```
katerinachinnappan, line 1: Pulling in @nina-py proposed schema changes

src/public/context.ts
```diff
@@ -1,24 +1,33 @@
...
+      request: Express.Request;
       db: PrismaClient;
     }
   ) {}
 
   get db(): IPublicContext['db'] {
     return this.config.db;
   }
+
+  get userId(): string {
```
katerinachinnappan, line 27: I think this is the way the `userId` will be passed to in the headers?

.circleci/config.yml
```diff
@@ -169,7 +169,7 @@ jobs:
             - node-v1-
       - run:
           name: install node modules
-          command: npm ci
+          command: npm install
```
kschelonka, line 172: 👍

schema-public.graphql
kschelonka: fyi - @hyperparabolic and @cmharlow plan to work on functions for de/serializing and include in apollo-utils

src/public/context.ts
kschelonka: You probably don't need to use this `get` pattern now that we're on apollo v4 and it's easy to pass request headers. I used this so that it was possible to mock the response from the context for some integration tests.

src/public/resolvers/mutations/ShareableList.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @param callback
+ */
+export async function executeMutation<T, U>(
+  context: IPublicContext,
+  data: T,
+  callback: (db, data: T, userId?: string) => Promise<U>
+): Promise<U> {
+  const { db, userId } = context;
+
+  if (!userId) {
```
kschelonka, line 22: I recommend moving this to the context or a specific model, so that you have finer control over your business logic over the domain (vs. re-exporting and wrapping all mutations).

src/public/resolvers/mutations/ShareableList.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  return entity;
+}
+
+/**
+ * @param parent
+ * @param data
+ * @param context
+ */
+export async function createShareableList(
```
kschelonka, line 36: Recommend consolidating over domain vs. operation -- so rather than a file with all mutations as a namespace, consider creating model for ShareableList entity and put the creation logic there (along with retrieval logic).

src/public/resolvers/mutations/ShareableList.ts
jpetto: i don't know if context will work, as this user check doesn't apply to every operation. all queries except one (for now) need this check, and all mutations need it. but agree this would be better moved to the model resolver layer.

(this pattern for mutations from collections is kind of a one-off that we haven't carried forward anywhere else. (we should probably standardize collections api.))

src/public/resolvers/mutations/ShareableList.ts
jpetto: yep, agree with kat. [curated-corpus-api](https://github.com/Pocket/curated-corpus-api/tree/main/src/admin/resolvers) has some nice structure in this regard.

jpetto (changes requested):
this is a great start on a sizeable PR 😄. a few requests, but nothing too big.

src/public/resolvers/mutations/ShareableList.ts
katerinachinnappan: not sure if I am following. What is the reason and what are the drawbacks of the current structure? @jpetto

.circleci/scripts/setup.sh
```diff
@@ -1,18 +1,21 @@
 #!/bin/bash
 
 set -e
+apt-get update && apt-get install -y sudo
 
 dir=$(dirname "$0")
+
+chmod +x "./"
```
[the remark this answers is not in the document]
katerinachinnappan: @jpetto unfortunately doesn't work without `chmod +x`. circleci complains about the scripts not being executable & then also about permissions on top of that. you can check some previous failed runs on this. 😢

.circleci/config.yml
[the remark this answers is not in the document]
katerinachinnappan: switched back to `ci` according to the slack thread on this. also removed caching step

schema-public.graphql
```diff
@@ -1,36 +1,59 @@
...
-	listItems: [ListItem!]!
+	status: ShareableListStatus!
+	moderationStatus: ShareableListModerationStatus!
+	createdAt: ISOString!
+	updatedAt: ISOString!
+	listItems: [ShareableListItem!]!
+}
+
+input CreateShareableListInput {
+	slug: String!
```
jpetto, line 44: a weekend of rest gives a clear mind - we won't get a slug when a list is created. slugs only need to be generated once the list is published, so we can drop this from the input 😄

(part of this is because, for MVP, slugs are immutable once generated, and will be based off the title. a user can edit a title as often as they want - so we want to delay creating the slug until we need it.)

src/public/context.ts
```diff
@@ -1,24 +1,33 @@
...
+    const userId = this.config.request.headers.userid;
+    return userId instanceof Array ? userId[0] : userId;
```
jpetto, line 29: is this sometimes an array?

src/public/resolvers/mutations/ShareableList.ts
jpetto: apologies for the ambiguity! i think i read @kat's suggestion wrong (but would be interested to hear more detail there).

my suggestion for this PR (and we may iterate on this pattern as we add more mutations), is to simply move the `userId` check (back?) into the `executeMutation` function. we know all mutations will require this check, so i think it's safe to perform this abstraction at this stage, and better to put here than in the database-specific function.

jpetto (changes requested):
hopefully cleared up some confusion i was a part of adding last week.

also remembered a slug will not be present at creation time, so we can drop all that input/logic 😄

src/public/context.ts
katerinachinnappan: @jpetto I believe `request.headers.userid` is actually an array.

src/database/types.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export type ShareableListItem = Omit<
+  ListItem,
+  | 'id'
+  | 'externalId'
+  | 'listId'
+  | 'itemId'
+  | 'sortOrder'
+  | 'createdAt'
+  | 'updatedAt'
```
jpetto, line 15: noting that @nina-py's PR has some changes here - but we can resolve those in her PR.

src/public/resolvers/mutations/ShareableList.integration.ts
```diff
@@ -0,0 +1,151 @@
...
+      };
+      const result = await request(app)
+        .post(graphQLUrl)
+        .set(headers)
+        .send({
+          query: print(CREATE_SHAREABLE_LIST),
+          variables: { data },
+        });
+      expect(result.body.data.createShareableList).not.to.exist;
+      expect(result.body.errors.length).to.equal(1);
```
jpetto, line 93: for a later PR - would be good to `expect` the error code (whatever `UserInputError` throws) here. it's a little more concrete/pertinent than the error message.

src/public/resolvers/mutations/ShareableList.integration.ts
```diff
@@ -0,0 +1,151 @@
...
+      );
+    });
+    it('should create List with existing title in db but for different userId', async () => {
+      const list1 = await createShareableListHelper(
+        db,
+        `Best Abstraction Art List`,
+        'wassilyKandinsky'
+      );
+      const title1 = list1.title;
+      // create new List with title1 value for the same user
```
jpetto, line 105: nit pick (non blocker): looks like a copy pasta comment that isn't accurate for this test 😄

src/public/resolvers/mutations/ShareableList.ts
```diff
@@ -0,0 +1,46 @@
...
+export async function executeMutation<T, U>(
+  context: IPublicContext,
+  data: T,
+  callback: (db, data: T, userId?: string) => Promise<U>
+): Promise<U> {
+  const { db, userId } = context;
+
+  if (!userId) {
+    throw new AuthenticationError('You must be logged in to use this service');
+  }
```
jpetto, line 24: one more thing for a follow-up PR: an integration test file (we've used the pattern of `ShareableList.auth.integration.ts`) where we validate the error thrown here.

jpetto (approved):
a couple small things that can be addressed in a follow up. (please make a ticket so we don't forget!) nice work on this big PR!

.circleci/scripts/setup.sh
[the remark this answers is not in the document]
efixler: @katerinachinnappan Have you tried `git update-index`? I know this is only supposed to be required when traversing windows file systems, but I have in the past had success using it to fix this issue where git doesn't persist file executable bits properly.

PR #32, opened 2023-02-14
OSL-197: updating userId from str to bigint, updating createShareable mutation to include listItems in response

katerinachinnappan:
## Goal

* Updated prisma schema -- `userId` from `String` to `Bigint`
* Created new prisma migration
* Updated `createShareableList` mutation to include `listItems` in the response
* Updated README

## Todos

- [x] deploy to dev

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-197](https://getpocket.atlassian.net/browse/OSL-197)


src/database/mutations/ShareableList.ts
```diff
@@ -13,7 +13,7 @@ export async function createShareableList(
 ): Promise<ShareableList> {
   // check if the title already exists for this user
   const titleExists = await db.list.count({
-    where: { title: data.title, userId: userId },
+    where: { title: data.title, userId: parseInt(userId) },
```
nina-py, line 16: Why don't we do the type conversion in the context? So that we don't have to do it over and over again in individual queries and mutations

src/public/context.ts
```diff
@@ -24,17 +25,13 @@ export class PublicContextManager implements IPublicContext {
     return this.config.db;
   }
 
-  get userId(): IPublicContext['userId'] {
+  get userId(): number | bigint {
```
nina-py, line 28: Should this change be made? I first thought maybe it was made for consistency, but `get db()` also returns a type picked from the `IPublicContext` interface, so makes sense to keep it?

src/database/mutations/ShareableList.ts
```diff
@@ -26,6 +26,9 @@ export async function createShareableList(
   }
 
   return db.list.create({
-    data: { ...data, userId },
+    data: { ...data, userId: userId },
```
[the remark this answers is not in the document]
katerinachinnappan: Ugh I think this sneaked in during the merge

nina-py (approved):
🔥

PR #35, opened 2023-02-14
OSL-198: removing `status` from `CreateShareableListInput`

katerinachinnappan:
## Goal

`Status` by default is always `PRIVATE` when creating a `List`, so removing `Status` from the body.

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-198](https://getpocket.atlassian.net/browse/OSL-198)


jpetto (changes requested):
just need to fix the linter and it'll be good to go!

PR #37, opened 2023-02-15
feat(mutation): [OSL-102] Add `updateShareableList` mutation

nina-py:
## Goal

Unblock the Web team by adding another mutation to the public graph.

- Added the mutation and integration tests.

- Updated `createShareableListHelper` to no longer generate slugs as they need to be set on publishing the list, not on initial list creation.

- Updated existing tests to cater for that.

- Refactored queries and mutations used in integration tests to use fragments to make sure each operation consistently returns all public values for lists and list items.

## I'd love to get feedback on

- Whether I've covered all edge cases both in code and tests.

## Tickets

- https://getpocket.atlassian.net/browse/OSL-102

### DIFF
--- a/schema-public.graphql
+++ b/schema-public.graphql
@@ -33,7 +33,7 @@ type ShareableListItem {
 
 type ShareableList {
 	externalId: String!
-	slug: String!
+	slug: String
 	title: String!
 	description: String
 	status: ShareableListStatus!
@@ -46,6 +46,13 @@ type ShareableList {
 input CreateShareableListInput {
 	title: String!
 	description: String
+}
+
+input UpdateShareableListInput {
+    externalId: String!
+    title: String
+    description: String
+    status: ShareableListStatus
 }
 
 type Query {
@@ -63,4 +70,8 @@ type Mutation {
 	Creates a Shareable List
 	"""
 	createShareableList(data: CreateShareableListInput!): ShareableList
+  """
+  Updates a Shareable List. This includes making it public.
+  """
+  updateShareableList(data: UpdateShareableListInput!): ShareableList!
 }
--- a/src/database/mutations/ShareableList.ts
+++ b/src/database/mutations/ShareableList.ts
@@ -1,6 +1,13 @@
-import { UserInputError } from '@pocket-tools/apollo-utils';
-import { PrismaClient } from '@prisma/client';
-import { CreateShareableListInput, ShareableList } from '../types';
+import { NotFoundError, UserInputError } from '@pocket-tools/apollo-utils';
+import { ListStatus, PrismaClient } from '@prisma/client';
+import slugify from 'slugify';
+import {
+  CreateShareableListInput,
+  ShareableList,
+  UpdateShareableListInput,
+} from '../types';
+import { getShareableList } from '../queries';
+import config from '../../config';
 
 /**
  * This mutation creates a shareable list, and _only_ a shareable list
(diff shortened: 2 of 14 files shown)


schema-public.graphql
```diff
@@ -33,7 +33,7 @@ type ShareableListItem {
 
 type ShareableList {
 	externalId: String!
-	slug: String!
+	slug: String
```
nina-py, line 36: Note I've made `slug` optional in the public schema now as it is indeed optional unless the user chooses to make their list public.

src/database/mutations/ShareableList.ts
```diff
@@ -32,3 +39,68 @@ export async function createShareableList(
...
+    const slugExists = await db.list.count({
+      where: { slug: data.slug, userId: userId },
+    });
+
+    if (slugExists) {
+      throw new UserInputError(
+        `A list with the slug "${data.slug}" already exists`
+      );
+    }
+  }
```
nina-py, line 97: Note I have added this check in as requested in the ticket, however I can't think of a way to write a test for it (maybe it's just been a very long day? 🤔 ). I can't force a slug to be anything it isn't to get a duplicate for the test given the checks we have around duplicate titles.

src/database/queries/ShareableList.ts
```diff
@@ -22,27 +22,8 @@ export async function getShareableList(
...
-          excerpt: true,
-          imageUrl: true,
-          authors: true,
-          sortOrder: true,
-          createdAt: true,
-          updatedAt: true,
-        },
-      },
+    include: {
+      listItems: true,
```
nina-py, line 26: I've updated this code to reduce potential maintenance as initially we were going to expose only a handful of properties of lists and list items on the graph but ended up exposing most of them.

src/database/mutations/ShareableList.ts
jpetto: yeah, this random number requirement makes testing difficult. also, i don't think we should throw an error if a slug already exists. clients can't really do anything to fix that error. i think we need to generate a unique slug.

what if we did the following:

- generate the slug based on the title without any random characters (what you already have on line 82)
- query the db for any slugs for the current user that start with the slug we just generated (e.g. `LIKE slug-we-generated%`)
- if there are results to that query, append a sequential number to the end of the slug we generated until we fail to match with the query results array (e.g. `best-recipes-cookies-3`)
- at this point, append the random characters

hopefully we could abstract the above a bit into a smaller function that could be unit tested.

not to blow up scope, but other thought here - should we store the random characters as a part of the slug? what if we used the first portion (8 chars before the first dash) of the `externalId` as the random sequence and computed the slug value we return to clients (saved `slug` + `-` + first chunk of `externalId`)? we don't have a hard rule that the characters have to be numeric (the characters don't serve a functional purpose - they are aesthetic 🙃). the external id is meant for public use anyway.

jpetto (dismissed):
we should not error on slug, but otherwise this looks great!

src/database/mutations/ShareableList.ts
katerinachinnappan: perhaps this is the where we should really think if we want to allow the user to edit the List title after the List has been published? If we disable title updates for a published list (and only allow description updates) then we don't need to worry about possible duplicate slugs?

nina-py: Thank you @jpetto! Summarising here the gist of the discussion we had in Slack around this:

- Let's save the slug as the slugified title in the database, no random numbers or alphanumerical characters appended.
- When serving the slug through the API, let's use the stored slug value + the first chunk of `externalId` (8 characters). 
- The above ^ will allow us to deliver a workable solution now and iterate on it later if need be.

src/public/resolvers/fieldResolvers.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const shareableListFieldResolvers = {
+  async slug(parent: ShareableList, _, { db, userId }): Promise<string> {
+    if (!parent.slug) {
+      // If a slug is not set, let's not touch it at all.
+      return null;
+    } else {
+      // Compose the slug value out of the stored value for it
+      // and a portion of the UUID string for `externalId`
+      const list = await getShareableList(db, userId, parent.externalId);
```
jpetto, line 15: would `parent` have the necessary values here? do we need the db call?

also could we add tests for this resolver?

src/public/resolvers/fieldResolvers.ts
```diff
@@ -0,0 +1,22 @@
...
+      return `${list.slug}-${list.externalId.substring(
+        0,
+        list.externalId.indexOf('-')
+      )}`;
```
jpetto, line 19: i know it's a small bit of functionality, but i wouldn't be mad if this was in a tiny helper function we could unit test 😄

src/database/mutations/ShareableList.ts
```diff
@@ -74,26 +74,12 @@ export async function updateShareableList(
...
 
   // If there is no slug and the list is being shared with the world,
-  // let's generate a slug from the title with a string of random numbers
-  // at the end.
+  // let's generate a slug from the title. Once set, it will not be
+  // updated to sync with any further title edits.
   if (data.status === ListStatus.PUBLIC && !list.slug) {
     // If an updated title is provided, generate the slug from that,
     // otherwise default to the title saved previously.
     data.slug = slugify(data.title ?? list.title, config.slugify);
```
jpetto, line 82: i think we still need dupe checks here, as we could get a collision if someone publishes list A, then edits the title of list A, then creates a new list B with the original title of list A, and then publishes list B.

in this case, i think we can simply append a numeric qualifier at the end of the generated slug, e.g. `best-cookie-recipes-3`.

jpetto: actually - i think you can ignore me here. appending a portion of the `externalId` should save us from this collision on the URL side. and i don't _think_ we need to care about duplicate slugs in the db...

src/public/resolvers/mutations/ShareableList.integration.ts
```diff
@@ -165,4 +184,243 @@ describe('public mutations: ShareableList', () => {
...
+
+      // Does the slug match the current title?
+      expect(updatedList.slug).to.contain(
+        slugify(updatedList.title, config.slugify)
+      );
+
+      // Does the slug contain the first 8 characters of the externalId?
+      expect(updatedList.slug).to.contain(
+        updatedList.externalId.substring(0, updatedList.externalId.indexOf('-'))
+      );
```
nina-py, line 330: Updated tests to look for the exact value of the first 8 characters of `externalId` in the value for `slug` served via a field resolver.

katerinachinnappan (approved):
🔥 🚀

PR #40, opened 2023-02-16
OSL-97: Adding shareableLists query & resolver & integration tests

katerinachinnappan:
## Goal

* Added the `shareableLists` query to the public schema (returns an array of Lists for a `userId` passed via the headers). This will unblock the web team and allow them to retrieve an array of Lists for a user.
    * Implemented the db query and `orderBy: { updatedAt: 'desc' }`
    * Implemented the resolver
    * Added integration tests
* Removed the temporary `lists` query and dummy data

## Todos

- [x] deploy to dev
- [x] wait for OSL-102 to be merged to update branch

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-97](https://getpocket.atlassian.net/browse/OSL-97)

### DIFF
--- a/src/database/queries/ShareableList.ts
+++ b/src/database/queries/ShareableList.ts
@@ -53,18 +53,7 @@ export async function getShareableLists(
       moderationStatus: true,
       createdAt: true,
       updatedAt: true,
-      listItems: {
-        select: {
-          url: true,
-          title: true,
-          excerpt: true,
-          imageUrl: true,
-          authors: true,
-          sortOrder: true,
-          createdAt: true,
-          updatedAt: true,
-        },
-      },
+      listItems: true,
     },
   });
 }
--- a/src/database/queries/ShareableList.ts
+++ b/src/database/queries/ShareableList.ts
@@ -44,15 +44,7 @@ export async function getShareableLists(
       moderationStatus: ModerationStatus.VISIBLE,
     },
     orderBy: { updatedAt: 'desc' },
-    select: {
-      externalId: true,
-      slug: true,
-      title: true,
-      description: true,
-      status: true,
-      moderationStatus: true,
-      createdAt: true,
-      updatedAt: true,
+    include: {
       listItems: true,
     },
   });
--- a/src/database/queries/ShareableList.ts
+++ b/src/database/queries/ShareableList.ts
@@ -46,3 +46,44 @@ export async function getShareableList(
     },
   });
 }
+
+/**
+ * Retrieves all available shareable lists for a given userId from the datastore.
+ *
+ * @param db
+ * @param userId
+ */
+export async function getShareableLists(
+  db: PrismaClient,
+  userId: number | bigint
+): Promise<ShareableList[]> {
+  return db.list.findMany({
+    where: {
+      userId,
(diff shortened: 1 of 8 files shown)


src/database/queries/ShareableList.ts
```diff
@@ -46,3 +46,44 @@ export async function getShareableList(
...
+          url: true,
+          title: true,
+          excerpt: true,
+          imageUrl: true,
+          authors: true,
+          sortOrder: true,
+          createdAt: true,
+          updatedAt: true,
+        },
+      },
```
nina-py, line 86: I think this can be shortened to 
```
include: {
   listItems: true
}
```
as in the individual query & both mutations now as gradually we added more and more props of `ShareableList` entity to the public graph and pulling in a handful we want to keep private, e.g. `moderatedBy`, will not be a significant effect on performance.

src/public/resolvers/queries/ShareableList.integration.ts
```diff
@@ -157,4 +165,129 @@ describe('public queries: ShareableList', () => {
...
+      expect(list1.description).to.equal(shareableList2.description);
+      expect(list1.status).to.equal(ListStatus.PRIVATE);
+      expect(list1.moderationStatus).to.equal(ModerationStatus.VISIBLE);
+      expect(list1.createdAt).not.to.be.empty;
+      expect(list1.updatedAt).not.to.be.empty;
+      expect(list1.externalId).not.to.be.empty;
+      // Empty list items array
+      expect(list1.listItems).to.have.lengthOf(0);
+
+      // Verify the second List props in the aray
```
nina-py, line 217: You could do a loop here over the `result.body.data.shareableLists` array and halve the number of `expect` statements. There's an example in the other query's tests: https://github.com/Pocket/shareable-lists-api/blob/ff954826f60a2cc31e7b8d65cb49a98389c86250/src/public/resolvers/queries/ShareableList.integration.ts#L149

src/public/resolvers/queries/sample-queries.gql.ts
```diff
@@ -25,6 +26,15 @@ export const GET_SHAREABLE_LIST = gql`
...
+export const GET_SHAREABLE_LISTS = gql`
+  query shareableLists {
+    shareableLists {
+      ...ShareableListPublicProps
+    }
+  }
+  ${ShareableListPublicProps}
+`;
+
 export const GET_LISTS = gql`
```
nina-py, line 38: I think this sample query can be removed now, too

nina-py (commented):
Great work! A few tweaks and it will be good to go! ⭐

src/database/queries/ShareableList.ts
```diff
@@ -27,3 +27,33 @@ export async function getShareableList(
...
+      externalId: true,
+      slug: true,
+      title: true,
+      description: true,
+      status: true,
+      moderationStatus: true,
+      createdAt: true,
+      updatedAt: true,
+      listItems: true,
+    },
```
nina-py, line 57: Sorry, I wasn't clear before. You can get rid of the entire `select` prop and replace it with `include: {listItems: true}`. This will fetch all props for lists and all props for the list items as well.

nina-py (approved):

PR #41, opened 2023-02-16
feat(mutation): [OSL-103] Add `addShareableListItem` mutation w/ tests

nina-py:
## Goal

Allow users to add stories (items) to their lists.

- [x] Added a new mutation and all checks 
- [x] Added tests

## Tickets

- https://getpocket.atlassian.net/browse/OSL-103


schema-public.graphql
```diff
@@ -21,7 +21,8 @@ enum ShareableListModerationStatus {
 }
 
 type ShareableListItem {
-	url: Url
+  externalId: ID!
+	url: Url!
```
nina-py, line 25: I've updated this type to add `externalId` which is an input for the `deleteShareableListItem` mutation, and I've also made `url` required/guaranteed as these are also unique identifiers and we compare them against other stories in the same list to see if there are any duplicates.

src/public/resolvers/mutations/ShareableListItem.integration.ts
```diff
@@ -41,20 +42,207 @@ describe('public mutations: ShareableListItem', () => {
...
+
+      const data: CreateShareableListItemInput = {
+        listExternalId: list.externalId,
+        url: 'https://www.test.com/duplicate-url',
+        sortOrder: 5,
+      };
+
+      const result = await request(app)
+        .post(graphQLUrl)
+        .send({
```
katerinachinnappan, line 162: seems like the header is missing from the request?

src/public/resolvers/mutations/ShareableListItem.integration.ts
```diff
@@ -41,20 +42,207 @@ describe('public mutations: ShareableListItem', () => {
...
+          variables: { data },
+        });
+
+      // There should be nothing in results
+      expect(result.body.data.createShareableListItem).to.be.null;
+
+      // And a "Forbidden" error
+      expect(result.body).to.have.nested.property(
+        'errors[0].extensions.code',
+        'FORBIDDEN'
```
katerinachinnappan, line 173: should we be testing for a `bad user input` error code instead? Seems like the `forbidden` code is thrown because the headers are missing from the request?

src/public/resolvers/mutations/ShareableListItem.integration.ts
```diff
@@ -41,20 +42,207 @@ describe('public mutations: ShareableListItem', () => {
...
+    it("should not create a new item for a list that doesn't exist", async () => {
+      const data: CreateShareableListItemInput = {
+        listExternalId: 'this-list-does-not-even-exist',
+        url: 'https://getpocket.com/discover',
+        sortOrder: 1,
+      };
+
+      const result = await request(app)
+        .post(graphQLUrl)
+        .send({
```
katerinachinnappan, line 67: headers missing in request?

src/public/resolvers/mutations/ShareableListItem.integration.ts
```diff
@@ -41,20 +42,207 @@ describe('public mutations: ShareableListItem', () => {
     await server.stop();
   });
 
-  beforeEach(async () => {
-    await clearDb(db);
-    // Create a VISIBLE List
-    shareableList = await createShareableListHelper(db, {
-      userId: parseInt(headers.userId),
-      title: 'Simon Le Bon List',
+  describe('createShareableListItem', () => {
```
katerinachinnappan, line 45: should we also add a test case for a user trying to add a list item to another user's list? The UI will probably not allow it but worth to check here?

katerinachinnappan (changes requested):
Looks great, just some minor comments about testing!

src/public/resolvers/mutations/ShareableListItem.integration.ts
nina-py: Well-spotted, thank you @katerinachinnappan !

src/public/resolvers/mutations/ShareableListItem.integration.ts
nina-py: Fixed!

katerinachinnappan (approved):
⭐ 🔥

PR #45, opened 2023-02-16
OSL-151: Adding deleteShareableListItem mutation, resolver & tests

katerinachinnappan:
## Goal

* Added `deleteShareableListItem` mutation to public schema along with the resolver and tests
    * A `ListItem` cannot be deleted if its parent `List` `moderationStatus` = `HIDDEN`

## Todos

- [ ] deploy to dev

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-151](https://getpocket.atlassian.net/browse/OSL-151)

### DIFF
--- a/src/database/mutations/ShareableListItem.ts
+++ b/src/database/mutations/ShareableListItem.ts
@@ -1,8 +1,4 @@
-import {
-  NotFoundError,
-  ForbiddenError,
-  UserInputError,
-} from '@pocket-tools/apollo-utils';
+import { NotFoundError, ForbiddenError } from '@pocket-tools/apollo-utils';
 import { ACCESS_DENIED_ERROR } from '../../shared/constants';
 import { PrismaClient } from '@prisma/client';
 import { ShareableListItem } from '../types';
@@ -19,13 +15,13 @@ export async function deleteShareableListItem(
   externalId: string,
   userId: number | bigint
 ): Promise<ShareableListItem> {
-  if (!externalId) {
-    throw new UserInputError('externalId must be provided.');
-  }
   // retrieve the existing ListItem before it is deleted
-  const listItem = await db.listItem.findFirst({
+  const listItem = await db.listItem.findUnique({
     where: {
       externalId,
+    },
+    include: {
+      list: true, // also retrieve the list
     },
   });
 
@@ -33,18 +29,14 @@ export async function deleteShareableListItem(
   if (!listItem) {
     throw new NotFoundError('A list item by that ID could not be found');
   }
-  // Retrieve the parent List of the ListItem to be deleted
-  const list = await db.list.findFirst({
-    where: {
-      userId,
-      listItems: {
-        some: { externalId: { in: [externalId] } },
-      },
-    },
-  });
+
+  // Check that the userId in the headers matches the userId of the List
+  if (Number(listItem.list.userId) !== userId) {
+    throw new ForbiddenError(ACCESS_DENIED_ERROR);
+  }
 
   // Check first the List moderation status. If HIDDEN, do not allow delete
-  if (list.moderationStatus == 'HIDDEN') {
+  if (listItem.list.moderationStatus == 'HIDDEN') {
     throw new ForbiddenError(ACCESS_DENIED_ERROR);
   }
   // delete ListItem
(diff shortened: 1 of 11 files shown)


src/database/mutations/ShareableListItem.ts
```diff
@@ -0,0 +1,57 @@
...
+ * @param userId
+ */
+export async function deleteShareableListItem(
+  db: PrismaClient,
+  externalId: string,
+  userId: number | bigint
+): Promise<ShareableListItem> {
+  if (!externalId) {
+    throw new UserInputError('externalId must be provided.');
+  }
```
nina-py, line 24: Can this error ever be thrown? I think GraphQL handles this as it won't allow the operation to run without required variables present, and you've made `externalId` required for that mutation.

src/public/resolvers/mutations/ShareableListItem.integration.ts
```diff
@@ -0,0 +1,161 @@
...
+        .send({
+          query: print(DELETE_SHAREABLE_LIST_ITEM),
+        });
+      expect(result.body.data).not.to.exist;
+      expect(result.body.errors.length).to.equal(1);
+      expect(result.body.errors[0].extensions.code).to.equal('BAD_USER_INPUT');
+      expect(result.body.errors[0].message).to.equal(
+        'Variable "$externalId" of required type "String!" was not provided.'
+      );
+    });
```
nina-py, line 72: We should not be testing standard GraphQL server functionality, so this test can be omitted entirely.

src/database/mutations/ShareableListItem.ts
```diff
@@ -0,0 +1,57 @@
...
+  // retrieve the existing ListItem before it is deleted
+  const listItem = await db.listItem.findFirst({
+    where: {
+      externalId,
+    },
+  });
```
nina-py, line 30: You can use `.findUnique` here instead of `.findFirst`. `.findFirst` is for cases where you're filtering the output by more than one field, but those fields include non-unique fields (e.g., for lists, it would be a combo of `externalId` and `moderationStatus`).

src/database/mutations/ShareableListItem.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+  // Retrieve the parent List of the ListItem to be deleted
+  const list = await db.list.findFirst({
+    where: {
+      userId,
+      listItems: {
+        some: { externalId: { in: [externalId] } },
+      },
+    },
+  });
```
nina-py, line 44: This is a bit of a round-about way of getting to the list. When you fetch the list item to be deleted from the DB, try 
```
include: { list }
```
as the relationship between these two entities should work both ways.

src/public/resolvers/mutations/ShareableListItem.integration.ts
```diff
@@ -0,0 +1,161 @@
...
+        });
+      expect(result.body.data).not.to.exist;
+      expect(result.body.errors.length).to.equal(1);
+      expect(result.body.errors[0].extensions.code).to.equal('NOT_FOUND');
+      expect(result.body.errors[0].message).to.equal(
+        'Error - Not Found: A list item by that ID could not be found'
+      );
+    });
+
+    it('should not delete a list item if parent list is hidden', async () => {
```
nina-py, line 105: Should we also test/cater for when somebody tries to remove an item from someone else's list?

nina-py (commented):
Solid effort! 🎉 

Some minor changes requested.

src/database/mutations/ShareableListItem.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  // Check if the ListItem was found
+  if (!listItem) {
+    throw new NotFoundError('A list item by that ID could not be found');
+  }
+
+  // Check that the userId in the headers matches the userId of the List
+  if (Number(listItem.list.userId) !== userId) {
```
katerinachinnappan, line 34: For some reason, prisma returns the id fields with the char `n` appended. I think this is because they are `BigInt` type? Just doing a `Number` cast here to ensure both fields are of the same type.

nina-py (approved):
🚀

PR #51, opened 2023-02-21
feat(mutation): [OSL-112] Add deleteShareableList mutation

efixler:
## Goal

Adding a delete API call for Shared Lists, and related tests.

## Todos

This PR implements the basic List delete. Deletes are currently "hard" I expect that these will need to become soft deletes when data retention policy is clear, and when opportune.

List Items will need to be deleted when a list is deleted (presumably) this is ticketed at [OSL-124](https://getpocket.atlassian.net/browse/OSL-214?atlOrigin=eyJpIjoiMzdlMTVmM2FhODlmNDIwMDliZDUzYzM2ZTgyYWE4MTkiLCJwIjoiamlyYS1zbGFjay1pbnQifQ) and will be address in a follow-on PR.

### Some Questions

- How are app metrics/observability handled? 
- How is input sanitization handled?

I didn't see any evidence of these in the codebase, so wondering where these are handled and how they work, to whatever extent they are materialized.

## Deployment

There should be no outstanding deployment issues here, there are no new migrations or secrets, etc.

## Tickets

[OSL-112 add deleteShareableList mutation](https://getpocket.atlassian.net/jira/software/c/projects/OSL/issues/OSL-112)

## Implementation Decisions

No major implementation decisions here. Deleting a list could be done in one query instead of two, but this depends on a prisma preview feature, and the implementation will change anyway (see todos)

[OSL-124]: https://getpocket.atlassian.net/browse/OSL-124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -1,6 +1,7 @@
 generator client {
   provider      = "prisma-client-js"
   binaryTargets = ["native", "debian-openssl-1.1.x"]
+  previewFeatures = ["extendedWhereUnique"]
 }
 
 datasource db {
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -1,7 +1,6 @@
 generator client {
   provider      = "prisma-client-js"
   binaryTargets = ["native", "debian-openssl-1.1.x"]
-  previewFeatures = ["extendedWhereUnique"]
 }
 
 datasource db {
--- a/src/database/mutations/ShareableList.ts
+++ b/src/database/mutations/ShareableList.ts
@@ -1,7 +1,6 @@
 import { NotFoundError, UserInputError } from '@pocket-tools/apollo-utils';
 import { ListStatus, PrismaClient } from '@prisma/client';
 import slugify from 'slugify';
-import { PrismaClientKnownRequestError } from '@prisma/client/runtime';
 import {
   CreateShareableListInput,
   ShareableList,
@@ -11,7 +10,7 @@ import {
 import { getShareableList } from '../queries';
 import config from '../../config';
 
-const PRISMA_RECORD_NOT_FOUND = "P2025";
+const PRISMA_RECORD_NOT_FOUND = 'P2025';
 
 /**
  * This mutation creates a shareable list, and _only_ a shareable list
@@ -96,9 +95,9 @@ export async function updateShareableList(
 }
 
 /**
- * This method deletes a shareable list. if the owner of the list 
- * represented by externalId matches the owner of the list. 
- * 
+ * This method deletes a shareable list. if the owner of the list
+ * represented by externalId matches the owner of the list.
+ *
  * @param db
  * @param externalId
  * @param userId
@@ -106,40 +105,42 @@ export async function updateShareableList(
 export async function deleteShareableList(
   db: PrismaClient,
   externalId: string,
-  userId: number | bigint 
+  userId: number | bigint
 ): Promise<ShareableList> {
   // Note for PR : input is unsanitized
   const deleteList = await db.list.findUnique({
-    where: { externalId: externalId } ,
+    where: { externalId: externalId },
     include: { listItems: true },
   });
   if (deleteList === null) {
     throw new NotFoundError(`List ${externalId} is not available`);
-  } else if (deleteList.userId !== BigInt(userId)) { 
+  } else if (deleteList.userId !== BigInt(userId)) {
     // local convention is to 404 when a normal user doesn't have resource ownership
     throw new NotFoundError(`List ${externalId} is not accessible`);
   }
-  
+
   // Now that we've checked that the we can delete the list, let's delete it.
   // We'll still make sure to existence errors for potential race conditions
-  // This operation is possible to execute with one query, using both the 
-  // externalId and the target userId but requires that the `extendedWhereUnique` 
+  // This operation is possible to execute with one query, using both the
+  // externalId and the target userId but requires that the `extendedWhereUnique`
   // prisma preview flag be enabled.
   // We don't need the return value here, since we have it above, so we
   // tell prisma to not select anything
-  await db.list.delete({
-    where: { externalId: externalId },
-    select: null,
-  }).catch((error) => {
-    // According to the Prisma docs, this should be a typed error 
-    // of type PrismaClientKnownRequestError, with a code, but it doesn't 
-    // come through typed
-    if (error.code === PRISMA_RECORD_NOT_FOUND) {
-      throw new NotFoundError(`List ${externalId} is not available`);
-    } else { // some unexpected DB error
-      throw(error);
-    }
-  });
+  await db.list
+    .delete({
+      where: { externalId: externalId },
+      select: null,
+    })
+    .catch((error) => {
+      // According to the Prisma docs, this should be a typed error
+      // of type PrismaClientKnownRequestError, with a code, but it doesn't
+      // come through typed
+      if (error.code === PRISMA_RECORD_NOT_FOUND) {
+        throw new NotFoundError(`List ${externalId} is not available`);
+      } else {
+        // some unexpected DB error
+        throw error;
+      }
+    });
   return deleteList;
 }
-
--- a/src/database/mutations/ShareableList.ts
+++ b/src/database/mutations/ShareableList.ts
@@ -107,7 +107,20 @@ export async function deleteShareableList(
   db: PrismaClient,
   externalId: string,
   userId: number | bigint 
-): Promise<String> {
+): Promise<ShareableList> {
+  const deleteList = await db.list.findUnique({
+    where: { externalId: externalId } ,
+    include: { listItems: true },
+  });
+  if (deleteList === null) {
+    throw new NotFoundError(`List ${externalId} is not available`);
+  } else if (deleteList.userId !== BigInt(userId)) { 
+    // local convention is to 404 when a normal user doesn't have resource ownership
+    throw new NotFoundError(`List ${externalId} is not accessible`);
+  }
+  return deleteList;
+  /*
+  
   // Note for PR : input is unsanitized
   
   // Since we 404 both in the case where the list doesn't exist, and in the case 
@@ -129,5 +142,6 @@ export async function deleteShareableList(
   
   console.log(deletedList);
   return deletedList.externalId;
+  */
 }
 
--- a/src/database/mutations/ShareableList.ts
+++ b/src/database/mutations/ShareableList.ts
@@ -108,6 +108,7 @@ export async function deleteShareableList(
   externalId: string,
   userId: number | bigint 
 ): Promise<ShareableList> {
+  // Note for PR : input is unsanitized
   const deleteList = await db.list.findUnique({
     where: { externalId: externalId } ,
     include: { listItems: true },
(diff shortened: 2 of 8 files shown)


prisma/schema.prisma
```diff
@@ -1,6 +1,7 @@
 generator client {
   provider      = "prisma-client-js"
   binaryTargets = ["native", "debian-openssl-1.1.x"]
+  previewFeatures = ["extendedWhereUnique"]
```
efixler, line 4: This will be deleted.

src/database/mutations/ShareableList.ts
```diff
@@ -1,14 +1,18 @@
 import { NotFoundError, UserInputError } from '@pocket-tools/apollo-utils';
 import { ListStatus, PrismaClient } from '@prisma/client';
 import slugify from 'slugify';
+import { PrismaClientKnownRequestError } from '@prisma/client/runtime';
```
efixler, line 4: Will be removed before merge. Leaving here for the PR in case anyone understands why these exceptions don't seem to propagate as expected.

src/database/mutations/ShareableList.ts
```diff
@@ -90,3 +94,52 @@ export async function updateShareableList(
...
+ * @param db
+ * @param externalId
+ * @param userId
+ */
+export async function deleteShareableList(
+  db: PrismaClient,
+  externalId: string,
+  userId: number | bigint 
+): Promise<ShareableList> {
+  // Note for PR : input is unsanitized
```
efixler, line 111: I noticed here (and generally across the service) that there seems to be no sanitization of text inputs. Where and how is that expected to happen?

katerinachinnappan: I believe this is done on the front-end side -- so the UI app.

nina-py: Some validation of inputs is done in the GraphQL layer - e.g. it will reject a string when a number is required. Prisma ORM does all input escaping for us.

efixler: Thanks @nina-py @katerinachinnappan. I'm going to do some investigation on this to understand our exposure wrt to user generated content on input, I'll ticket that separately.

Any other issues/questions with the review? I'd like to merge today, if possible :)

src/database/mutations/ShareableList.ts
```diff
@@ -90,3 +93,54 @@ export async function updateShareableList(
...
+  const deleteList = await db.list.findUnique({
+    where: { externalId: externalId },
+    include: { listItems: true },
+  });
+  if (deleteList === null) {
+    throw new NotFoundError(`List ${externalId} is not available`);
+  } else if (deleteList.userId !== BigInt(userId)) {
+    // local convention is to 404 when a normal user doesn't have resource ownership
+    throw new NotFoundError(`List ${externalId} is not accessible`);
```
katerinachinnappan, line 119: this should probably be a `ForbiddenError` as the list does exists but we don't allow access if its a different user.

src/public/resolvers/mutations/ShareableList.integration.ts
```diff
@@ -185,6 +186,63 @@ describe('public mutations: ShareableList', () => {
...
+      const result = await request(app)
+        .post(graphQLUrl)
+        .set(headers)
+        .send({
+          query: print(DELETE_SHAREABLE_LIST),
+          variables: { externalId: otherUserList.externalId },
+        });
+      expect(result.body.data).to.be.null;
+      expect(result.body.errors.length).to.equal(1);
+      expect(result.body.errors[0].extensions.code).to.equal('NOT_FOUND');
```
katerinachinnappan, line 205: would check for `forbiddenError` instead

src/database/mutations/ShareableList.ts
efixler: I had asked about this in channel and @kschelonka indicated (IIRC) that at Pocket we 404 when the user doesn't have permission to view a resource. Either pattern (403 or 404) is acceptable, as long as we are consistent internally about the rules. (And whatever our rule is for this case, it should be documented _somewhere_)

src/database/mutations/ShareableList.ts
```diff
@@ -90,3 +93,54 @@ export async function updateShareableList(
...
+  await db.list
+    .delete({
+      where: { externalId: externalId },
+      select: null,
+    })
+    .catch((error) => {
+      // According to the Prisma docs, this should be a typed error
+      // of type PrismaClientKnownRequestError, with a code, but it doesn't
+      // come through typed
+      if (error.code === PRISMA_RECORD_NOT_FOUND) {
```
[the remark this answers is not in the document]
efixler: If the prisma error here is uncaught it will then bubble up to the client and the error we emit will not be correct -- the service will return an error with the code `INTERNAL_SERVER_ERROR`, which is not what we want. 

The case is rare, but it's still important to be defensive and race-aware when modifying the db, and sending the right response to clients so they are getting the error they expect for the case, in this case the `NOT_FOUND` error. 

In part, this is working around Prisma's behavior -- Prisma is considering a delete that deletes zero rows to be an error, which is arguable. 

In any event, what we're doing here is ensuring that the method completes as expected, and ensuring that the backend's behavior is normal from the client's perspective.

src/database/mutations/ShareableList.ts
```diff
@@ -90,3 +93,54 @@ export async function updateShareableList(
...
+  });
+  if (deleteList === null) {
+    throw new NotFoundError(`List ${externalId} is not available`);
+  } else if (deleteList.userId !== BigInt(userId)) {
+    // local convention is to 404 when a normal user doesn't have resource ownership
+    throw new NotFoundError(`List ${externalId} is not accessible`);
+  }
+
+  // Now that we've checked that the we can delete the list, let's delete it.
+  // We'll still make sure to existence errors for potential race conditions
```
nina-py, line 123: Nitpick: "existence"?

src/database/mutations/ShareableList.ts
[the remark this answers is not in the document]
nina-py: Agree that it's a rare but possible case. To be thorough, should our test coverage include a test for this edge case? 
And we can then follow up with another ticket to update the other delete mutation to follow the same pattern and possibly abstract some of that code into a helper function.

src/database/mutations/ShareableList.ts
```diff
@@ -90,3 +93,54 @@ export async function updateShareableList(
...
+  // This operation is possible to execute with one query, using both the
+  // externalId and the target userId but requires that the `extendedWhereUnique`
+  // prisma preview flag be enabled.
+  // We don't need the return value here, since we have it above, so we
+  // tell prisma to not select anything
+  await db.list
+    .delete({
+      where: { externalId: externalId },
+      select: null,
+    })
```
nina-py, line 133: I haven't seen `select: null` used before so I was curious and turned on query logging to compare if it makes a difference. Here's what I found. Note that this is just the delete operation and doesn't include the initial lookup query to find the list to delete.

With `select: null`:
```
      prisma:query BEGIN
      prisma:query SELECT `shareablelists`.`List`.`id`, `shareablelists`.`List`.`externalId`, `shareablelists`.`List`.`userId`, `shareablelists`.`List`.`slug`, `shareablelists`.`List`.`title`, `shareablelists`.`List`.`description`, `shareablelists`.`List`.`status`, `shareablelists`.`List`.`moderationStatus`, `shareablelists`.`List`.`moderatedBy`, `shareablelists`.`List`.`moderationReason`, `shareablelists`.`List`.`createdAt`, `shareablelists`.`List`.`updatedAt` FROM `shareablelists`.`List` WHERE (`shareablelists`.`List`.`externalId` = ? AND 1=1) LIMIT ? OFFSET ?
      prisma:query SELECT `shareablelists`.`List`.`id` FROM `shareablelists`.`List` WHERE (`shareablelists`.`List`.`externalId` = ? AND 1=1)
      prisma:query DELETE FROM `shareablelists`.`List` WHERE (`shareablelists`.`List`.`id` IN (?) AND (`shareablelists`.`List`.`externalId` = ? AND 1=1))
      prisma:query COMMIT
```

Without the `select` prop (omitting it entirely):
```
      prisma:query BEGIN
      prisma:query SELECT `shareablelists`.`List`.`id`, `shareablelists`.`List`.`externalId`, `shareablelists`.`List`.`userId`, `shareablelists`.`List`.`slug`, `shareablelists`.`List`.`title`, `shareablelists`.`List`.`description`, `shareablelists`.`List`.`status`, `shareablelists`.`List`.`moderationStatus`, `shareablelists`.`List`.`moderatedBy`, `shareablelists`.`List`.`moderationReason`, `shareablelists`.`List`.`createdAt`, `shareablelists`.`List`.`updatedAt` FROM `shareablelists`.`List` WHERE (`shareablelists`.`List`.`externalId` = ? AND 1=1) LIMIT ? OFFSET ?
      prisma:query SELECT `shareablelists`.`List`.`id` FROM `shareablelists`.`List` WHERE (`shareablelists`.`List`.`externalId` = ? AND 1=1)
      prisma:query DELETE FROM `shareablelists`.`List` WHERE (`shareablelists`.`List`.`id` IN (?) AND (`shareablelists`.`List`.`externalId` = ? AND 1=1))
      prisma:query COMMIT
```

Verdict: no difference. 

Prisma, however, as many other ORMs, tends to select all fields by default when making a query, so how can this be improved? I reintroduced the `select` prop but only selected one field:
```
select: { id: true }
```
and got the following queries:
```
      prisma:query BEGIN
      prisma:query SELECT `shareablelists`.`List`.`id` FROM `shareablelists`.`List` WHERE (`shareablelists`.`List`.`externalId` = ? AND 1=1) LIMIT ? OFFSET ?
      prisma:query SELECT `shareablelists`.`List`.`id` FROM `shareablelists`.`List` WHERE (`shareablelists`.`List`.`externalId` = ? AND 1=1)
      prisma:query DELETE FROM `shareablelists`.`List` WHERE (`shareablelists`.`List`.`id` IN (?) AND (`shareablelists`.`List`.`externalId` = ? AND 1=1))
      prisma:query COMMIT
```
So the latter at least shortens the additional lookup query, but unfortunately there doesn't seem to be a way to eliminate it as unnecessary overhead. It's good to know that Prisma does run an extra check for the entry's existence before deleting it.

If wanting to optimise for performance, you could delete the list by its primary key (`id`) instead of the `VARCHAR` `externalId`, since you already know its value from the lookup at the start of the DB resolver function. If leaving the `where` condition as-is, it can be shortened here thanks to TypeScript to `where: { externalId }`.

nina-py (changes requested):

src/database/mutations/ShareableList.ts
kschelonka: @nina-py I think the inconsistency here comes from the access pattern. For everything list-related we don't have a single ID, so everything has to be accessed using user-id as essentially part of a compound key. That means that we don't really model an authZ error when it comes to accessing specific items.

efixler: Thanks @kschelonka @nina-py to reiterate, either pattern is acceptable, presuming that our rules are consistent across the board, e.g. across all services. @kschelonka maybe I'm missing some nuance in your explanation but I don't believe that the deciding factor pertains to the key structure for a resource, it pertains to whether we want to expose the existence of resources to un-permissioned users. It is not unheard of for APIs to take the not-found approach in this case. 

All that said, I'm generally a believer in sticking to simple plain access patterns, and, unless there's a strong precedent for the 404 approach, the 403 approach is more in keeping with the spirit and expectations of the web. Furthermore, I don't think there's material harm in acknowledging that the resource actually does exist, so I think we concur on the outcome here.

src/database/mutations/ShareableList.ts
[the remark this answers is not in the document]
efixler: @nina-py : Thanks. The pattern here _is_ correct, and we should be applying similar semantics to similar DB cases, since we're not running (nor should we be) transactionally, and what prisma considers an error in this case really isn't. All it takes is an errant script or a [malicious or non-malicious] traffic surge to surface these sorts of race conditions.

However, I do not believe that this case is testable under our current "integration" testing patterns (at least not without modifying to method for the sole purpose of enabling testing, which is not a good approach generally).

Ideally, we'd be able to mock the prisma client and monkeypatch method responses when testing, which would enable us to do execute decoupled, "real" unit testing. This is going to be a potentially important piece of testing infrastructure for us to build out, but it's not trivial. I'd like to continue that conversation and figure out how to get us to a place where we can execute meaningful unit tests, faster and with less dependencies than we have right now.

So, let's skip the test right now. Not ideal, but the codeflow and results are hand-tested. @katerinachinnappan happy to work through the semantics for list item deletion.

src/database/mutations/ShareableList.ts
```diff
@@ -90,3 +93,54 @@ export async function updateShareableList(
...
+    include: { listItems: true },
+  });
+  if (deleteList === null) {
+    throw new NotFoundError(`List ${externalId} is not available`);
+  } else if (deleteList.userId !== BigInt(userId)) {
+    // local convention is to 404 when a normal user doesn't have resource ownership
+    throw new NotFoundError(`List ${externalId} is not accessible`);
+  }
+
+  // Now that we've checked that the we can delete the list, let's delete it.
```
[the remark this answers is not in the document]
efixler: Umm, yeah, but appreciate the close reading :)

src/database/mutations/ShareableList.ts
efixler: Good catch @nina-py ! All I did was read the prisma docs, which suggested that passing null would skip deserialization. (We can discuss prisma's queries offline, but my take on these is that there's a fair bit of fluff. All that said query logging FTW :) )

efixler: Committed following suggestion.

efixler:
@nina-py : Addressed all PR feedback, thanks (see inline comment wrt to test for race condition, going to skip that here). 

Unless there's anything else, let's wrap this up!

src/database/mutations/ShareableList.ts
kschelonka: can explain a little more in slack, but for saves there isn't a concept of an authorization error for accessing a specific resource because the userId is part of the compound key required for lookup. there's no such thing as a resource owned by another user -- it's a UserResource essentially. We can't differentiate between a 404 and 403 in this case.

katerinachinnappan (approved):

nina-py (approved):

PR #58, opened 2023-02-23
chore(resolvers): [OSL-208] Use `ISOString` scalar resolver from `apollo-utils`

nina-py:
## Goal

Keep the new repo up with the latest changes. Updated both public and admin resolvers.

## Tickets

https://getpocket.atlassian.net/browse/OSL-208


katerinachinnappan (approved):
✨⭐️

PR #60, opened 2023-02-23
feat(mutation): [OSL-214] Delete list items when deleting a list

efixler:
## Goal

Delete list items when deleting a list.

This PR is pretty straightforward, but it did raise questions wrt Pocket's usage of foreign keys, commented inside.

## Tickets

- Link to JIRA tickets

[OSL-214](https://getpocket.atlassian.net/browse/OSL-214)

## Implementation Decisions

Followed existing patterns. Would have preferred to not make list deletion subject to list item deletion but this is forced by key constraints at present. Also, we could arguably use to have integration tests directly against the mutations. There were not there at present, so I skipped this and tested on the public route where `deleteAllListItemsForList` is used instead. 

[OSL-214]: https://getpocket.atlassian.net/browse/OSL-214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/database/mutations/ShareableList.ts
+++ b/src/database/mutations/ShareableList.ts
@@ -10,6 +10,7 @@ import {
   ShareableList,
   UpdateShareableListInput,
 } from '../types';
+import { deleteAllListItemsForList } from './ShareableListItem';
 import {
   ACCESS_DENIED_ERROR,
   PRISMA_RECORD_NOT_FOUND,
@@ -122,6 +123,23 @@ export async function deleteShareableList(
   } else if (deleteList.userId !== BigInt(userId)) {
     throw new ForbiddenError(ACCESS_DENIED_ERROR);
   }
+  // This delete must occur before the list is actually deleted,
+  // due to a foreign key constraint on ListIitems. We should remove this
+  // foreign key constraint for a number of reasons.
+  // In the small context here of deletes, the foreign key constraint makes
+  // this action both less safe and slower:
+  // Less safe: If the list item deletion encounters a failure partway through
+  // the list will remain in place, and will also be mangled. In this case it
+  // would be far preferable to guarantee deletion of the list entity -- thus removing
+  // it from userspace -- first. It is not critical that the list item table be
+  // consistent; there is no harm (other than disk space) in orphaned list item rows,
+  // and we're eventually going to need to clean things up anyway.
+  // Slower: Without the foreign key constraint, we do not need to `await` the result
+  // of the list item deletion, it could happen asynchronously
+  // For these and similar reasons foreign keys are typically not used in environments
+  // running at high scale (e.g. Etsy, etc)
+  // Leaving this in now so we can discuss and circle back and keep moving :)
+  await deleteAllListItemsForList(db, deleteList.id);
 
   // Now that we've checked that we can delete the list, let's delete it.
   // We'll catch the case where the list has been deleted under us, to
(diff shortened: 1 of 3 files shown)


src/database/mutations/ShareableList.ts
```diff
@@ -122,6 +123,23 @@ export async function deleteShareableList(
...
+  // due to a foreign key constraint on ListIitems. We should remove this
+  // foreign key constraint for a number of reasons.
+  // In the small context here of deletes, the foreign key constraint makes
+  // this action both less safe and slower:
+  // Less safe: If the list item deletion encounters a failure partway through
+  // the list will remain in place, and will also be mangled. In this case it
+  // would be far preferable to guarantee deletion of the list entity -- thus removing
+  // it from userspace -- first. It is not critical that the list item table be
+  // consistent; there is no harm (other than disk space) in orphaned list item rows,
+  // and we're eventually going to need to clean things up anyway.
```
kschelonka, line 136: this makes sense to me

jpetto (approved):

src/database/mutations/ShareableList.ts
nina-py: My vote is for leaving the foreign constraint in. Why leave things to the application layer if it can be enforced so easily on the database layer?

nina-py (approved):

src/database/mutations/ShareableList.ts
efixler: Folks: <3 for having these conversations in context and for making these small PRs fun and interesting. Everything is connected :) 

Coincidentally, a related topic came up in Steve's Fireside Chat yesterday: In the document he shared about M&A he called out characteristics in running Cloud Services at scale, and specifically mentioned "eventually-consistent" (I can't link to the specific paragraph so here's a [screenshot](https://p201.p0.n0.cdn.getcloudapp.com/items/mXuD8wvZ/6ebe3cbd-cbb8-42bd-b6ee-f42e9e41f934.png?v=5f46e797295b373d172b4c5bdd954b0c) and [link to doc](https://docs.google.com/document/d/1hSspIi0wulQLaLURSovL-C_0B9nxo45iCDeF0fu5SZk/edit?usp=sharing) . What is generally meant here is that systems running with high scale characteristics need to balance consistency and performance needs in a way that considers the requirements of the particular subdomain (wikipedia)[https://en.wikipedia.org/wiki/Eventual_consistency]. 

I don't think we're today, and I think we probably don't need to be there for this early phase of Shared Lists, but I also believe that it's helpful for us to visit these issues as they come up so we can illuminate them and mark out how we'll approach fundamental patterns moving forward. 

As far as the data store goes, experience has generally led me to accept that a well-indexed DB where the important indexes are memory-resident, with everything else as simple and unconstrained as possible, is a strong functional approach to keeping the datastore performant and malleable moving forward. There's definitely lots of capabilities that are convenient, but they tend to get problematic at scale. (The use of autoincrement numeric ids and prisma also constrains some approaches to scaling the database moving forward so there's that)

All that said, ^^ will [hopefully] be evolutions we will need to work on together as this feature launches and grows. At this moment, let's stick with the approach, and I'm going to merge this PR as-is. I would like us to flag any relations like this in the future because I do believe it's going to be important to be mindful about them in higher stress areas than this one.  And, to @jpetto's point/question, let's keep important technical conversations alive, keep moving along the current path and not get blocked, and not dive into things that might be dangerous :) 

Again, grateful for the conversation and collaboration and really excited to get this out to people :)

PR #61, opened 2023-02-23
OSL-224: adding itemId as query input param for createShareableList mutation

katerinachinnappan:
## Goal

* Added `itemId` as optional input param for `createShareableList` mutation
* Added `itemId` to `ShareableListItem` type in public schema

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-224](https://getpocket.atlassian.net/browse/OSL-224)

### DIFF
--- a/schema-public.graphql
+++ b/schema-public.graphql
@@ -78,7 +78,7 @@ type Mutation {
 	"""
 	createShareableList(data: CreateShareableListInput!): ShareableList
 	"""
-	Deletes a Shareable List
+	Deletes a Shareable List.
 	"""
 	deleteShareableList(externalId: ID!): ShareableList!
 	"""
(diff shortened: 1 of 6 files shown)


schema-public.graphql
```diff
@@ -47,13 +48,14 @@ input UpdateShareableListInput {
...
-  authors: String
-  sortOrder: Int!
+	listExternalId: ID!
+	itemId: Int
+	url: Url!
+	title: String
+	excerpt: String
+	imageUrl: Url
+	authors: String
+	sortOrder: Int!
```
nina-py, line 58: All good with regards to code but I'm just curious is that tabs vs. spaces here that is producing a large diff or there's actually no difference, just git being glitchy?

nina-py (approved):

schema-public.graphql
katerinachinnappan: @nina-py good catch, not git being glitchy but I had my indentation set to 4, fixed!

nina-py (approved):

PR #62, opened 2023-02-24
feat(items): [OSL-225] remove code adding externalId characters to the slug

nina-py:
## Goal

Simplify the code!

- Removed field resolvers for the `slug` field and updated tests.

- Verified that slug field uniqueness is guaranteed per user with the DB unique constraint of `userId` + `slug`.

## Tickets

https://getpocket.atlassian.net/browse/OSL-225


jpetto (approved):
deletions & simplification!

PR #64, opened 2023-02-25
OSL-228: ensuring generated slug is unique

katerinachinnappan:
## Goal

* Ensuring generated slug for a list first time made public is unique. Adding a consecutive number to generated slug based on # of slugs containing list title in db.
* Added edge test case.

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-228](https://getpocket.atlassian.net/browse/OSL-228)

### DIFF
--- a/src/database/mutations/ShareableList.ts
+++ b/src/database/mutations/ShareableList.ts
@@ -83,12 +83,26 @@ export async function updateShareableList(
   }
 
   // If there is no slug and the list is being shared with the world,
-  // let's generate a slug from the title. Once set, it will not be
+  // let's generate a unique slug from the title. Once set, it will not be
   // updated to sync with any further title edits.
   if (data.status === ListStatus.PUBLIC && !list.slug) {
-    // If an updated title is provided, generate the slug from that,
-    // otherwise default to the title saved previously.
-    data.slug = slugify(data.title ?? list.title, config.slugify);
+    // First check how many slugs containing the list title already exist in the db
+    const slugCount = await db.list.count({
+      where: {
+        slug: { contains: slugify(data.title ?? list.title, config.slugify) },
+      },
+    });
+    // if there is at least 1 slug containing title of list to update,
+    // append next consecutive # of slugCount to data.slug
+    if (slugCount) {
+      data.slug =
+        slugify(data.title ?? list.title, config.slugify) +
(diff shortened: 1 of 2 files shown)


katerinachinnappan:
I went with a simple approach of just appending the next consecutive int to a generated slug. 

So we have an entry in the db: 

1. A list with the slug `hello-world` but its title has been updated to `Bye World`
2. User wants to create a new list with title `Hello World` and makes it `PUBLIC`, the generated slug would be `hello-world-2`.

Question: can we ever get access to the `userName`? I know we pass the `userId` in the headers, but if we can get the `userName` as well, then we could easily generate a unique slug: `kchinnappan-hello-world` and `jpetto-hello-world` and never have to worry about appending anything? `userName` is not sensitive data and it is unique.

nina-py:
I don't think the API will be getting anything but `userId`. Also, the complete URLs for public lists are going to have the format of `getpocket.com/@username/slug`, so having the username in the URL twice would probably not look that good.

nina-py (approved):
🚀! Excellent comments!

src/database/mutations/ShareableList.ts
```diff
@@ -83,12 +83,26 @@ export async function updateShareableList(
...
   // updated to sync with any further title edits.
   if (data.status === ListStatus.PUBLIC && !list.slug) {
-    // If an updated title is provided, generate the slug from that,
-    // otherwise default to the title saved previously.
-    data.slug = slugify(data.title ?? list.title, config.slugify);
+    // First check how many slugs containing the list title already exist in the db
+    const slugCount = await db.list.count({
+      where: {
+        slug: { contains: slugify(data.title ?? list.title, config.slugify) },
+      },
```
jpetto, line 93: apologies if it wasn't clear from the ticket, but slugs only need to be unique _per user_. so the above `where` clause should include `userId`. i.e. both you and i could have a public list with the slug `best-cafes`, but i cannot have two public lists with the slug `best-cafes`.

(probably needs one more test added to clarify this functionality.)

jpetto: one other question here (that also doesn't need to block as this works!) - is there an advantage to using `contains` instead of `startsWith`?

jpetto (commented):
one bug here that we should fix, and a couple questions 😄.

PR #65, opened 2023-02-27
chore(docs): [OSL-205] Document GraphQL schemas thoroughly

nina-py:
## Goal

Document our schemas to the required standard (example: https://github.com/Pocket/curated-corpus-api/blob/main/schema-admin.graphql) 

## Tickets

https://getpocket.atlassian.net/browse/OSL-205


efixler (approved):
🔥

PR #68, opened 2023-02-27
OSL-125-OSL-230: adding searchShareableList query to admin graph

katerinachinnappan:
## Goal

* Adding `searchShareableList` query to the admin graph. Only query param is `externalId`
* Adding a `PrismaBigIntResolver` for the `id` and `userId` fields of the `ShareableList` type on the admin graph.
* Adding integration tests (including auth tests) for `searchShareableList` query.
* Ensuring slugs are generated uniquely per `userId` (also added a test case for this).
* Updating README on how to authorize admin graph operations.

## Todos

- [x] deploy to dev

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-125](https://getpocket.atlassian.net/browse/OSL-125)
- [https://getpocket.atlassian.net/browse/OSL-230](https://getpocket.atlassian.net/browse/OSL-230)


schema-admin.graphql
```diff
@@ -6,7 +6,7 @@ See Section 5.6 of the RFC 3339 profile of the ISO 8601 standard: https://www.ie
 scalar ISOString
 
 type ShareableList {
-  id: ID!
+  id: Float!
```
nina-py, line 9: I noticed this while documenting the schema - I don't think we need to expose the internal, integer ID of the shareable list even on the Admin schema. We'll never query or use it as one of the inputs, `externalId` is used instead for all entities in Curation Admin Tools.

src/admin/resolvers/index.ts
```diff
@@ -0,0 +1,12 @@
+import { PocketDefaultScalars } from '@pocket-tools/apollo-utils';
+import { searchShareableList } from './queries/ShareableList';
+import { PrismaBigIntResolver } from './fields/PrismaBigInt';
+
+export const resolvers = {
+  ...PocketDefaultScalars,
+  ShareableList: {
+    id: PrismaBigIntResolver,
+    userId: PrismaBigIntResolver,
+  },
```
nina-py, line 10: Given that the new `PrismaBigIntResolver` resolves the `userId` field, should it be placed in some sort of a shared directory and used in the public schema as well instead of appearing only on the Admin schema? Thinking of a pathname, something like `src/shared/fields/`?

src/admin/resolvers/fragments.gql.ts
```diff
@@ -0,0 +1,22 @@
+import { gql } from 'graphql-tag';
+
+/**
+ * This GraphQL fragment contains all the properties that must be available
+ * in the Admin Pocket Graph for a Shareable List.
+ */
+export const ShareableListPublicProps = gql`
+  fragment ShareableListPublicProps on ShareableList {
+    id
```
nina-py, line 9: Should the `id` property be removed from the type returned and this sample query as it should not be exposed by the API?

nina-py (commented):
Looking very good 👍 , a few minor changes requested

src/admin/resolvers/fields/PrismaBigInt.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @constructor
+ */
+export const PrismaBigIntResolver = (parent, args, context, info) => {
+  return parseFieldToInt(parent[info.fieldName]);
+};
+
+function parseFieldToInt(field: string) {
+  if (!field) {
+    return null;
+  }
```
efixler, line 19: Should we guard here for the case where `parseInt` returns `NaN`?

src/admin/resolvers/queries/ShareableList.auth.integration.ts
```diff
@@ -0,0 +1,122 @@
...
+      expect(result.body.data.searchShareableList).to.be.null;
+      expect(result.body.errors[0].message).to.equal(ACCESS_DENIED_ERROR);
+      expect(result.body.errors[0].extensions.code).to.equal('FORBIDDEN');
+    });
+
+    it('should fail if user does not have access', async () => {
+      const badHeaders = {
+        name: 'Test User',
+        username: 'test.user@test.com',
+        // missing any collection/readoly group
```
efixler, line 63: speling (readoly, sorry :) )

src/admin/resolvers/index.ts
jpetto: as we're never going to expose `id` to clients, we also shouldn't need the `id` resolver here, right?

src/admin/resolvers/queries/ShareableList.integration.ts
```diff
@@ -0,0 +1,127 @@
...
+      expect(list.moderationStatus).to.equal(ModerationStatus.VISIBLE);
+
+      // Variable values that just need to be non-null - we know Prisma
+      // returns them in a compatible format
+      expect(list.createdAt).not.to.be.empty;
+      expect(list.updatedAt).not.to.be.empty;
+      expect(list.externalId).not.to.be.empty;
+
+      // Does the slug match the list title?
+      expect(list.slug).to.equal(slugify(shareableList.title, config.slugify));
```
jpetto, line 124: nit - we can probably just do a `not.to.be.empty` check here, as we verify the `slug` value in other tests. i appreciate the thoroughness though! 😄

src/database/mutations/ShareableList.ts
```diff
@@ -89,6 +89,7 @@ export async function updateShareableList(
     // First check how many slugs containing the list title already exist in the db
     const slugCount = await db.list.count({
       where: {
+        userId,
         slug: { contains: slugify(data.title ?? list.title, config.slugify) },
```
jpetto, line 93: not a blocker or bug, but `startsWith` instead of `contains` might be a bit more efficient.

src/database/types.ts
```diff
@@ -17,6 +17,11 @@ export type ShareableList = Omit<
   listItems?: ShareableListItem[];
 };
 
+/**
+ * This is the shape of a shareable list object on the admin Pocket Graph.
+ */
+export type ShareableListComplete = Omit<List, 'listItems'>;
```
jpetto, line 23: as this is the type we'll use on the admin side, i think it's better for it to contain _all_ data for a list _except_ `id`.

we can retrieve the list items from the db as we do for the public resolver, but (for now) not add them to the graphql schema. (graphql will ignore any data points present in an object that aren't defined in the graphql schema.). i can actually see showing the list items in the admin tool being useful to moderators.

src/public/resolvers/mutations/ShareableList.integration.ts
```diff
@@ -545,6 +545,79 @@ describe('public mutations: ShareableList', () => {
       expect(updatedList2.slug).to.equal('hangover-hotel-2');
     });
 
+    it('should generate two identical slugs but for two different users', async () => {
```
jpetto, line 548: nice!

src/test/helpers/updateShareableListHelper.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export async function updateShareableListHelper(
+  prisma: PrismaClient,
+  externalId: string,
+  data: UpdateListHelperInput
+): Promise<List> {
+  const input: UpdateListHelperInput = {
+    status: data.status,
+    slug: data.slug,
+  };
```
jpetto, line 23: is this object declaration necessary? could you just use the `data` argument that was passed in?

jpetto (changes requested):
a few small things to be updated, but overall looking good!

src/admin/resolvers/index.ts
katerinachinnappan: yep, just userId

src/test/helpers/updateShareableListHelper.ts
katerinachinnappan: this is to prevent data from being `any` type as we don't have a  `UpdateListHelperInput` defined in types and we are passing a hard-coded slug

nina-py (approved):
LGTM! 🚀

jpetto (approved):

PR #70, opened 2023-02-28
feat(list items): [OSL-240] Add `publisher` field to List Items

nina-py:
## Goal

Unblock the frontend as they need to save and display the `publisher` field on stories added to shareable lists.

- Added `publisher` field to Prisma schema.

- Added `publisher` to list of props on the `ShareableListItem` type in the public schema and to `CreateShareableListItemInput` to take in that value from the frontend when a new list item is created.

- Updated tests and test helpers to test for the presence of this field everywhere list items are manipulated or returned.

## Tickets

https://getpocket.atlassian.net/browse/OSL-240


katerinachinnappan (approved):
🔥

jpetto (commented):
🔥

PR #73, opened 2023-03-01
feat(queries): [OSL-147, OSL-149] Set up a `shareableListPublic` query

nina-py:
## Goal

Set up a query for publicly available lists. 

- [x] Moved the basic resolver scaffolding from the [closed PR](https://github.com/Pocket/shareable-lists-api/pull/54).
- [x] Added a 403 error for lists that have been taken down.
- [x] Updated `userId` resolver to point to `PrismaBigInt` field resolver as it does on the admin schema.
- [x] Context update - we don't need a logged-in Pocket user for this query
- [x] Added integration test coverage
- [x] Updated the query to take in `externalId` as the parameter to find lists by and addressed remaining review feedback.

## Tickets

- https://getpocket.atlassian.net/browse/OSL-147
- https://getpocket.atlassian.net/browse/OSL-149


src/public/resolvers/queries/ShareableList.integration.ts
```diff
@@ -166,6 +170,114 @@ describe('public queries: ShareableList', () => {
...
+      });
+
+      // Run the query we're testing
+      const result = await request(app)
+        .post(graphQLUrl)
+        .send({
+          query: print(GET_SHAREABLE_LIST_PUBLIC),
+          variables: {
+            slug: list.slug,
+            userId: Number(list.userId),
```
efixler, line 215: @nina-py : This change (and the similar one below line 245) are what made the tests pass. (I had a little time and I was both curious and wanted to return the favor of merging my last PR : ) 

Not sure if you want to keep if this way, but the tests pass now.

nina-py: Yay! Thank you @efixler!

schema-public.graphql
```diff
@@ -147,12 +147,15 @@ input CreateShareableListItemInput {
...
+	"""
+	shareableList(externalId: ID!): ShareableList
   """
-  Looks up and returns a Shareable List with a given external ID for a given user.
-  (the user ID will be coming through with the headers)
+  Returns a publicly-shared Shareable List. Note: this query does not require user authentication.
   """
-  shareableList(externalId: ID!): ShareableList
-
+  shareableListPublic(slug: String!, userId: Float!): ShareableList
```
jpetto, line 158: based on slack conversations and the current unreliability of `username`, we're going to fall back to just accepting the list's `externalId`. urls get uglier, but code is simpler!

src/public/resolvers/utils.ts
```diff
@@ -39,3 +43,18 @@ export function sanitizeMutationInput<InputType>(input: InputType): InputType {
...
+ *
+ * @param userId
+ */
+export function validateUserId(userId: string | string[]): number | bigint {
+  // We need this check for nearly every query and mutation on the public graph
+  if (!userId) {
+    throw new ForbiddenError(ACCESS_DENIED_ERROR);
+  }
+
+  return userId instanceof Array ? parseInt(userId[0]) : parseInt(userId);
```
jpetto, line 59: is there any reason not to keep this line of code in `context.ts`? seems this could just be checked once at request time...

jpetto (changes requested):
one (more) change on query inputs based on slack convos - at least the code gets simpler! and one question on `userId` validation.

schema-public.graphql
nina-py: Hooray! I think this is what this long-suffering PR started with :)

src/public/resolvers/utils.ts
nina-py: Good point, moving it back to public context!

jpetto (approved):
🚀

PR #76, opened 2023-03-01
OSL-245: enforce list title and description length during list update

katerinachinnappan:
## Goal

Enforce list title and description length during list update. Updated tests.

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-245](https://getpocket.atlassian.net/browse/OSL-245)


kschelonka:
FYI - You can use the `@constraint` directive to annotate your schema directly with these validations.

kschelonka (approved):

PR #78, opened 2023-03-02
feat(sanitization): [OSL-227] add XSS protection to all public mutations

nina-py:
## Goal

Sanitize all inputs! 

- Added the `xss` package.

- Added a new util function that sanitises inputs & added usage of it through the `executeMutation` function.

- Added a new GitHub workflow to run a check over unit tests.

- Modified existing integration tests to add checks for sanitization.

## Tickets

- https://getpocket.atlassian.net/browse/OSL-227

### DIFF
--- a/src/public/resolvers/mutations/utils.spec.ts
+++ b/src/public/resolvers/mutations/utils.spec.ts
@@ -1,0 +1,66 @@
+import { expect } from 'chai';
+import {
+  CreateShareableListInput,
+  CreateShareableListItemInput,
+} from '../../../database/types';
+import { sanitizeMutationInput } from './utils';
+
+describe('utility functions', () => {
+  describe('xssifyMutationInput', () => {
+    it('transforms strings in a mutation input object', () => {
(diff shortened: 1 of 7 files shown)


src/public/resolvers/mutations/utils.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+        'Trying out this new Pocket feature&lt;script&gt;alert("!!!");&lt;/script&gt;'
+      );
+    });
+
+    it('returns numeric values as-is in a mutation input object', () => {
+      const input: CreateShareableListItemInput = {
+        listExternalId: '123-abc',
+        itemId: 456789,
+        url: 'https://www.test.com/story',
+        title: 'This is a test title',
```
jpetto, line 30: might be nice to test an emoji character here (i checked locally and it does preserve it).

jpetto (approved):
nice and tight! only question here is - are we moving away from circle for running unit tests? if not, could we continue that pattern here (instead of the github workflow)?

PR #79, opened 2023-03-02
refactor(errors): use NOT_FOUND in more places for security

jpetto:
## Goal

use `NOT_FOUND` error to obfuscate potential malicious API request responses. (better if an attacker doesn't know that the list / list item ID they are maliciously trying to manipulate is valid.)

## Tickets

- https://getpocket.atlassian.net/browse/OSL-206

### DIFF
--- a/src/database/mutations/ShareableListItem.ts
+++ b/src/database/mutations/ShareableListItem.ts
@@ -1,14 +1,7 @@
-import {
-  ForbiddenError,
-  NotFoundError,
-  UserInputError,
-} from '@pocket-tools/apollo-utils';
+import { NotFoundError, UserInputError } from '@pocket-tools/apollo-utils';
 import { ModerationStatus, PrismaClient } from '@prisma/client';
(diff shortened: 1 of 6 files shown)


src/database/mutations/ShareableListItem.ts
```diff
@@ -39,14 +42,9 @@ export async function createShareableListItem(
...
-  // Check that the userId in the headers matches the userId of the List
-  if (Number(list.userId) !== userId) {
-    throw new ForbiddenError(ACCESS_DENIED_ERROR);
-  }
-
   // check if an item with this URL already exists in this list
-  const itemExists = await db.listItem.count({
-    where: { listId: list.id, url: data.url },
+  const itemExists = list.listItems.find((item) => {
+    return item.url === data.url;
```
jpetto, line 47: i _think_ requesting list items above is more performant than another db call here. happy to be proven wrong though!

PR #80, opened 2023-03-02
feat(mutation): [OSL-126] Mutation to moderate shareable lists

efixler:
## Goal

Add the ability to moderate shared lists. 

This PR only covers the admin mutations, e.g. setting the moderation data. 

Restricting viewing of moderated list. etc will happen in [OSL-249](https://getpocket.atlassian.net/browse/OSL-249) . I split these in part for general size, and because the consumption of moderation status is, at it turns out, not required rn.

## Reference

Documentation here:

- Link to docs

## Tickets

- [OSL-126](https://getpocket.atlassian.net/browse/OSL-126) 

## Implementation Decisions

This is (or at least should be :) generally following the boilerplates. 

[OSL-249]: https://getpocket.atlassian.net/browse/OSL-249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OSL-126]: https://getpocket.atlassian.net/browse/OSL-126?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


README.md
```diff
@@ -48,7 +48,7 @@ The sample headers below allow full access to all queries and mutations:
 {
   "groups": "mozilliansorg_pocket_moderation_full",
   "name": "Matt McPockets",
-  "username": "ad|Mozilla-LDAP|mmcpockets"
+  "username": "mmcpockets"
```
efixler, line 51: Based on a conversation in Slack I think we expect only the `username` here, so I updated the docs

src/admin/resolvers/mutations/ShareableList.ts
```diff
@@ -0,0 +1,37 @@
...
+  parent,
+  { data },
+  context: IAdminContext
+): Promise<ShareableListComplete> {
+  const { db, authenticatedUser } = context;
+  if (!authenticatedUser.hasFullAccess) {
+    throw new ForbiddenError(ACCESS_DENIED_ERROR);
+  }
+  data.moderatedBy = authenticatedUser.username;
+  data.moderationReason = data.moderationReason.trim();
```
efixler, line 32: There didn't seem to be an explicit requirement that `moderationReason` be non-empty, but it seemed reasonable to expect the moderator to leave a comment as to why they did what they did. I can remove this if needed.

nina-py (approved):
Looks great! I think whether making the moderation reason comment required or not is something we'll figure out during the pilot - it's fine to make it required for now.

PR #81, opened 2023-03-02
feat(pilot user db model): add prisma model for pilot user

jpetto:
## Goal

create a prisma model for pilot users.

## Tickets

- https://getpocket.atlassian.net/browse/OSL-247

## Implementation Decisions

- kept as simple as possible with a slight eye to the future (`mozillaEmployee` and `createdBy` fields).


nina-py (approved):
🚀

PR #82, opened 2023-03-03
OSL-145: Emit snowplow events  to shared event-bridge

katerinachinnappan:
## Goal

Emit snowplow events from shareable lists api to shared event-bridge.

* Added `src/snowplow` directory to hold logic for emitting snowplow events.
* Updated README on Snowplow
* Updated CircleCI config to run unit tests 
* Updated mutations to send the payload to the event-bridge.
* Installed `sinon` JS lib for stubbing AWS EventClientBridge in tests.

## Todos

- [x] deploy to dev

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-145](https://getpocket.atlassian.net/browse/OSL-145
)


.circleci/config.yml
```diff
@@ -215,10 +215,9 @@ jobs:
 workflows:
   all:
     jobs:
-      # UNCOMMENT WHEN ADDING FIRST UNIT TEST
-      # - pocket/node_mocha_ts_test:
-      #     <<: *not_main
-      #     name: test_specs
+      - pocket/node_mocha_ts_test:
+          <<: *not_main
+          name: test_specs
```
nina-py, line 220: We have a GitHub action to run unit tests, does it not encompass these new tests?

README.md
```diff
@@ -125,3 +125,82 @@ If tests don't rely on other services:
...
+This API sends two kinds of events to the Pocket Event Bridge --> Snowplow: `shareable-list` and `shareable-list-item` events.
+
+The following are event types defined for the `shareable-list` event:
+
+- `shareable-list-created`
+- `shareable-list-updated`
+- `shareable-list-deleted`
+- `shareable-list-published`
+- `shareable-list-unpublished`
+- `shareable-list-hidden`
```
nina-py, line 140: I would send people off to check the schema in Snowplow and/or the spec repository. Keeping another copy of it in the docs in sync with everything else will be hard going forward.

src/database/mutations/ShareableList.ts
```diff
@@ -116,14 +126,40 @@ export async function updateShareableList(
...
+      isShareableListEventType: true,
+    });
+  }
+  // if only list title or description are updated, send event bridge event for shareable-list-updated event type
+  if (data.title || data.description) {
+    await sendEvent(EventBridgeEventType.SHAREABLE_LIST_UPDATED, {
+      shareableList: updatedList as ShareableListComplete,
+      isShareableListEventType: true,
+    });
+  }
```
nina-py, line 160: Are we sending the update event regardless of sending the publish/unpublish event? The frontend will probably have three different buttons for it so it's not an issue but there's potential here to send publish+update or unpublish+update events

src/database/mutations/ShareableListItem.ts
```diff
@@ -58,9 +61,19 @@ export async function createShareableListItem(
     listId: list.id,
   };
 
-  return db.listItem.create({
+  const listItem = await db.listItem.create({
     data: input,
   });
```
nina-py, line 66: This function is typed to return a promise, should it be updated after this change? Or will that work fine as is?

nina-py (commented):
A few questions but otherwise solid effort! ⭐

src/snowplow/types.ts
```diff
@@ -0,0 +1,65 @@
...
+  shareableListItemExternalId?: string;
+  listExternalId?: string;
+  isShareableListEventType?: boolean;
+  isShareableListItemEventType?: boolean;
+}
+
+/**
+ * This ShareableList type maps to the shareable_list entity defined in Snowplow
+ */
+export type SnowplowShareableList = {
```
sri-kirsh, line 37: probably at this point, we only have snowplow as the consumer for the events

but in-general we don't tie event emission to a consumer and emit as much domain related events and whatever format make sense for the microservice

and if the consumer wants it it in a specific format, then the consuming repo can do the ETL e.g [here](https://github.com/Pocket/shared-snowplow-consumer/blob/main/src/eventConsumer/userEvents/userEventConsumer.ts#L42)

 this is specifically useful if we see multiple consumers in the future (e.g another microservice as a consumer)

README.md
sri-kirsh: nice idea documenting the models. 

+1 on deciding if it should live in consumption repo as its consumer specific.. emitters can document its domain/business events

src/snowplow/types.ts
katerinachinnappan: @sri-kirsh good point! I did tho intend the `snowplow` directory to be targeted for only snowplow events, hence I did the transformation in the repo.

src/database/mutations/ShareableListItem.ts
katerinachinnappan: @nina-py it seems to work fine!

README.md
katerinachinnappan: good point, I will just refer to the schema definition in the spec repository.

src/database/mutations/ShareableList.ts
katerinachinnappan: @nina-py good catch. do we want to only send `shareable-list-updated` events for only title/description updates?

.circleci/config.yml
katerinachinnappan: @nina-py I think that is only for main branch ? This runs tests on other branches, i don't have a preference if we should keep it

src/database/mutations/ShareableList.ts
katerinachinnappan: Updated the condition to check only for list title or description updates.

src/database/mutations/ShareableListItem.ts
jpetto: afaik `async` functions _always_ return a promise.

src/database/mutations/ShareableListItem.ts
```diff
@@ -58,9 +61,19 @@ export async function createShareableListItem(
...
+
+  //send event bridge event for shareable-list-item-created event type
+  await sendEvent(EventBridgeEventType.SHAREABLE_LIST_ITEM_CREATED, {
+    shareableListItem: listItem,
+    shareableListItemExternalId: listItem.externalId,
+    listExternalId: list.externalId,
+    isShareableListItemEventType: true,
+  });
```
jpetto, line 74: do we need to `await` these calls? would be nice if we didn't have to block clients waiting for the event bridge call. i think we can _not_ await and it'll complete in the node event loop? maybe?

src/database/mutations/ShareableList.ts
```diff
@@ -116,14 +126,40 @@ export async function updateShareableList(
...
+  // if only list title or description are updated, send event bridge event for shareable-list-updated event type
+  if ((data.title || data.description) && !data.status) {
+    await sendEvent(EventBridgeEventType.SHAREABLE_LIST_UPDATED, {
+      shareableList: updatedList as ShareableListComplete,
+      isShareableListEventType: true,
+    });
+  }
```
jpetto, line 160: as this is a fair amount of event code, what do you think about factoring this out into its own function?

src/database/mutations/ShareableList.ts
```diff
@@ -116,14 +126,40 @@ export async function updateShareableList(
...
     include: {
       listItems: true,
     },
   });
+
+  // this mutation does a lot of things so we need to break down the operations to determine
+  // what events to send to snowplow
+
+  // if list was published, send event bridge event for shareable-list-published event type
+  if (data.status === ListStatus.PUBLIC) {
```
jpetto, line 141: something to watch out for here - the web client _may_ pass in the current value of any property. so i think we need to first check if `data.status !== list.status`.

jpetto (changes requested):
a few areas to address, but this is a really good start!

src/database/mutations/ShareableListItem.ts
katerinachinnappan: i think if we don't use `await` it won't resolve into the `shareableListComplete` type?

jpetto: apologies for the ambiguity! i meant the `sendEvent` calls specifically. i think we could let those calls complete on their own time and not block returning of data to the client.

src/database/mutations/ShareableList.ts
```diff
@@ -258,3 +286,39 @@ function shareableListTitleDescriptionValidation(
...
+      isShareableListEventType: true,
+    });
+  }
+  // if list was unpublished, send event bridge event for shareable-list-unpublished event type
+  if (data.status !== list.status && data.status === ListStatus.PRIVATE) {
+    sendEvent(EventBridgeEventType.SHAREABLE_LIST_UNPUBLISHED, {
+      shareableList: updatedList as ShareableListComplete,
+      isShareableListEventType: true,
+    });
+  }
```
jpetto, line 312: nit pick - this could be simplified a little bit by first checking if `data.status !== list.status`, then in an inner `if/else if` check the status change (as only one status can be true per request).

basically, as it stands now, if the first `if` condition is met, attempting to run the second `if` condition is just using cpu cycles for no reason 😄

src/database/mutations/ShareableList.ts
```diff
@@ -258,3 +286,39 @@ function shareableListTitleDescriptionValidation(
...
+  // if only list title or description are updated, send event bridge event for shareable-list-updated event type
+  if (
+    ((data.title && data.title !== list.title) ||
+      (data.description && data.description !== list.description)) &&
+    !data.status
```
jpetto, line 317: i think it would be good to check web-client patterns here. the _might_ send all data here, even if it hasn't changed, meaning the `!data.status` check above could be faulty. what is the intent of the `!data.status` check? is there a reason not to send the `SHAREABLE_LIST_UPDATED` event if all three properties (`title`, `description`, and `status`) are present in `data`?

src/database/mutations/ShareableList.ts
```diff
@@ -258,3 +286,39 @@ function shareableListTitleDescriptionValidation(
     );
   }
 }
+
+/**
+ * updateShareableList mutation does a lot of things so we need to break down the operations in a helper function
+ * to determine what events to send to snowplow
+ **/
+function sendEventHelper(
```
jpetto, line 294: just because there's another `sendEventHelper` defined in another file, maybe rename this to be more specific? maybe like `updateShareableListSendEventHelper`? then you wouldn't have to rename the import from `snowplow`.

jpetto (changes requested):
a few more small tweaks, but it's very close 😄

src/public/resolvers/mutations/ShareableList.integration.ts
```diff
@@ -39,6 +41,7 @@ describe('public mutations: ShareableList', () => {
   let server: ApolloServer<IPublicContext>;
   let graphQLUrl: string;
   let db: PrismaClient;
+  let s3Stub: sinon.SinonStub;
```
jpetto, line 44: nit - this variable name (also in the other `integration.ts` file) is a bit misleading, as we aren't stubbing `S3` 😉

src/snowplow/events.ts
```diff
@@ -0,0 +1,221 @@
...
+  SnowplowShareableListItem,
+  ShareableListEventBusPayload,
+  ShareableListItemEventBusPayload,
+} from './types';
+
+/**
+ * This function takes in the API Shareable List object and transforms it into a Snowplow Shareable List object
+ * @param shareableList
+ */
+async function transformAPIShareableListToSnowplowShareableList(
```
jpetto, line 22: do these transformation functions need to be `async`?

src/snowplow/events.ts
```diff
@@ -0,0 +1,221 @@
...
+  };
+}
+
+/**
+ * This method sets up the payload to send to Event Bridge for shareable-list mutations
+ *
+ * @param eventType
+ * @param shareableList
+ */
+export async function generateShareableListEventBridgePayload(
```
jpetto, line 85: same question - do these generate functions need to be `async`?

jpetto (changes requested):
a _tiny_ bit more cleanup 😄

src/public/resolvers/mutations/ShareableList.integration.ts
katerinachinnappan: good catch, i took this from my test in reparser 😅

jpetto (approved):
yay! ship it!

PR #84, opened 2023-03-03
feat(isPilotUser query): query to know if user is in pilot program

jpetto:
## Goal

add a basic query to know if the `userId` passed in the header is a part of the pilot program.

- removed unnecessary async declarations

## Tickets

- https://getpocket.atlassian.net/browse/OSL-248

## Implementation Decisions

removed `async` declarations on functions that have no `await` calls. (i don't think this has any performance impact either way, but it's not needed.)


katerinachinnappan (approved):
⭐

PR #86, opened 2023-03-03
fix(list title update): sending existing title should not fail

jpetto:
## Goal

fix a bug where sending the list's current title value to `updateShareableList` would cause a "duplicate title" error.

## Reference

- [Slack thread](https://pocket.slack.com/archives/C04HWMUKLN5/p1677865852780239)

## Tickets

- https://getpocket.atlassian.net/browse/OSL-251


katerinachinnappan (approved):
good catch!! 🔥

PR #88, opened 2023-03-06
OSL-243: add listItemData param to createShareableList mutation

katerinachinnappan:
## Goal

According to t[his slack thread](https://pocket.slack.com/archives/C04HWMUKLN5/p1677691820523829), we want to have the option to create a `ShareableListItem` when creating a `ShareableList`. Updating the `createShareableList` mutation to take in an optional `listItemData` param.

## Todos

- [ ] deploy to dev

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-243](https://getpocket.atlassian.net/browse/OSL-243)

### DIFF
--- a/schema-public.graphql
+++ b/schema-public.graphql
@@ -135,7 +135,11 @@ input UpdateShareableListInput {
 Input data for creating a Shareable List Item.
 """
 input CreateShareableListItemInput {
-  listExternalId: ID!
+  """
+  listExternalId parameter is optional as it is not
+  required when creating a ShareableList with a ShareableListItem
+  """
+  listExternalId: ID
   itemId: Float
   url: Url!
   title: String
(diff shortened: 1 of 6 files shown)


schema-public.graphql
```diff
@@ -135,7 +135,11 @@ input UpdateShareableListInput {
 Input data for creating a Shareable List Item.
 """
 input CreateShareableListItemInput {
-  listExternalId: ID!
+  """
+  listExternalId parameter is optional as it is not
+  required when creating a ShareableList with a ShareableListItem
+  """
+  listExternalId: ID
```
nina-py, line 142: This change exposes the API to requests where list items are added without specifying a valid list using the `CreateShareableListItem` mutation. I would create another input that does not contain `listExternalId` and that is used in the `CreateShareableList` mutation only instead of reusing an existing input for the list item mutation.

nina-py: Can we omit this parameter altogether? I don't think it's ever going to be used, is that right?

schema-public.graphql
```diff
@@ -146,6 +146,25 @@ input CreateShareableListItemInput {
   sortOrder: Int!
 }
 
+"""
+Input data for creating a Shareable List Item during Shareable List creation.
+"""
+input CreateShareableListItemWithShareableListInput {
```
nina-py, line 152: Trying to think of a shorter, more semantic name for this input... and failing miserably! 🙃 

What about `CreateShareableListItemAlongsideNewListInput`?

nina-py (approved):
🔥

PR #89, opened 2023-03-07
fix(mutation): [OSL-255] retrieving userId from updateShareableList mutation should not throw error

nina-py:
## Goal

Fix a bug + improve test coverage so that it doesn't happen again.

- Updated sample fragment for lists to retrieve `userId` so that it has a full list of returned fields.

- Updated tests to assert `userId` is returned as a result of the update mutation.

## Tickets

- https://getpocket.atlassian.net/browse/OSL-255

### DIFF
--- a/src/public/resolvers/mutations/ShareableList.integration.ts
+++ b/src/public/resolvers/mutations/ShareableList.integration.ts
@@ -372,6 +372,7 @@ describe('public mutations: ShareableList', () => {
         listToUpdate.createdAt.toISOString()
       );
       expect(updatedList.listItems).to.have.lengthOf(0);
+      expect(updatedList.userId).to.equal(parseInt(headers.userId));
 
       // The `updatedAt` timestamp should change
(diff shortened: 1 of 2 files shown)


src/public/resolvers/mutations/ShareableList.integration.ts
```diff
@@ -519,7 +520,7 @@ describe('public mutations: ShareableList', () => {
       expect(updatedList.slug).to.equal(slugify(data.title, config.slugify));
     });
 
-    it('should append next consectuive number to generated slug for list made public for the first time if slug contaning list title already exists for a single user', async () => {
+    it('should append next consecutive number to generated slug for list made public for the first time if slug containing list title already exists for a single user', async () => {
```
nina-py, line 523: Fixed a couple of unrelated typos in the comments as my IDE highlighted them.

jpetto (approved):

PR #90, opened 2023-03-07
chore(validation): [OSL-246] add @constraint directive to enforce max length on fields

nina-py:
## Goal

WIP - `graphql-constraint-directive` package doesn't work as expected.

## Tickets

- https://getpocket.atlassian.net/browse/OSL-246

### DIFF
--- a/schema-public.graphql
+++ b/schema-public.graphql
@@ -85,11 +85,11 @@ type ShareableList {
   """
   The title of the list. Provided by the Pocket user.
   """
-  title: String!
+  title: String! @constraint(min: 3, max: 100)
   """
   Optional text description of a Shareable List. Provided by the Pocket user.
   """
-  description: String
+  description: String @constraint(min: 3, max: 200)
   """
   The status of the list. Defaults to PRIVATE.
   """
@@ -126,8 +126,8 @@ Input data for updating a Shareable List.
 """
 input UpdateShareableListInput {
   externalId: ID!
-  title: String
-  description: String
+  title: String @constraint (min: 3, max: 100)
+  description: String @constraint (min: 3, max: 200)
   status: ShareableListStatus
 }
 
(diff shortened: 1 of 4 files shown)


src/public/server.ts
```diff
@@ -23,13 +28,26 @@ export function getPublicServer(
   httpServer: Server
 ): ApolloServer<IPublicContext> {
   const cache = getRedisCache();
+
+  // Add @constraint directive to the schema
+  const schema = buildSubgraphSchema([
+    { typeDefs: gql(constraintDirectiveTypeDefs) },
+    {
+      typeDefs: typeDefsPublic,
+      resolvers,
```
kschelonka, line 37: Not 100% sure, but if the link directive stuff from slack doesn't solve the compilation issue all the way, try compiling typedefs into one array

```suggestion
    { typeDefs: [gql(constraintDirectiveTypeDefs), typeDefsPublic],
      resolvers,
```

schema-public.graphql
kschelonka, line 1: Need to add @link and @composeDirective

https://pocket.slack.com/archives/C034FB364LC/p1679430084165889?thread_ts=1679378488.120289&cid=C034FB364LC

nina-py:
Notes:

Adding `@composeDirective` into existing directives declaration results in 
```
"INVALID_LINK_DIRECTIVE_USAGE
[shareable-lists-api] Cannot import unknown element "@composeDirective"
```
^^^ which is reasonable, given the existing declaration is for v2.0 and `@composeDirective` is introduced in v2.1

Updating the existing declaration to v2.1 results in 
```
INVALID_GRAPHQL
[shareable-lists-api] Unknown directive "@composeDirective".

UNKNOWN_FEDERATION_LINK_VERSION
[shareable-lists-api] Invalid version v2.1 for the federation feature in @link direction on schema
```
and further cascading errors.

jpetto (approved):
i know this took a long time, but really nice work! i like how clean this is - even down to building the schemas in code rather than in circle.

should we file a follow-up ticket to remove the app code that validates content length?

PR #93, opened 2023-03-08
feat(API authorization): validate pilot user on necessary API calls

jpetto:
## Goal

validate the `userId` in headers is in the pilot user group. applies to all public API endpoints _except_ `shareableListPublic`.

- standardize integration test error checking

## Tickets

- https://getpocket.atlassian.net/browse/OSL-196

### DIFF
--- a/src/public/resolvers/mutations/ShareableList.integration.ts
+++ b/src/public/resolvers/mutations/ShareableList.integration.ts
@@ -320,6 +320,27 @@ describe('public mutations: ShareableList', () => {
   });
 
   describe('deleteShareableList', () => {
+    it('should not delete a list for a non-pilot user', async () => {
+      const theList = await createShareableListHelper(db, {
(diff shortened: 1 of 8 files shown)


src/public/resolvers/mutations/ShareableList.integration.ts
```diff
@@ -7,6 +7,7 @@ import {
   List,
```
katerinachinnappan, line 7: should we also add a test to check that only pilot users can update/delete a list? I see there is this check for creating a list

katerinachinnappan (changes requested):

src/public/resolvers/mutations/ShareableList.integration.ts
jpetto: yes! good catch. adding now.

katerinachinnappan (approved):

PR #99, opened 2023-03-13
OSL-263: shareableListPublic query should not return PRIVATE list

katerinachinnappan:
## Goal

This [slack thread](https://pocket.slack.com/archives/C04HWMUKLN5/p1678725002822039) informs that the `shareableListPublic` query returns both `PRIVATE` and `PUBLIC` list. This PR updates the query to return only a `PUBLIC` list.

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-263](https://getpocket.atlassian.net/browse/OSL-263)

### DIFF
--- a/src/database/queries/ShareableList.ts
+++ b/src/database/queries/ShareableList.ts
@@ -42,6 +42,8 @@ export async function getShareableListPublic(
   db: PrismaClient,
   externalId: string
 ): Promise<ShareableList> {
+  // externalId is unique, but the generated type for `findUnique` here doesn't
+  // include `status`, so using `findFirst` instead
   const list = await db.list.findFirst({
     where: {
       externalId,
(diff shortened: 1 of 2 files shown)


src/database/queries/ShareableList.ts
```diff
@@ -42,9 +42,10 @@ export async function getShareableListPublic(
   db: PrismaClient,
   externalId: string
 ): Promise<ShareableList> {
-  const list = await db.list.findUnique({
+  const list = await db.list.findFirst({
```
Herraj, line 45: Any specific reason we're using `findFirst` here?

src/public/resolvers/queries/ShareableList.integration.ts
```diff
@@ -250,11 +250,39 @@ describe('public queries: ShareableList', () => {
...
+        moderationStatus: ModerationStatus.VISIBLE,
+      });
+
+      // Run the query we're testing
+      const result = await request(app)
+        .post(graphQLUrl)
+        .send({
+          query: print(GET_SHAREABLE_LIST_PUBLIC),
+          variables: {
+            externalId: newList.externalId,
```
Herraj, line 267: nit pick: would be slightly more readable if we were to name this variable `privateList` 😅

Herraj (approved):
🚀

PR #104, opened 2023-03-14
feat(admin graph): [OSL-267] return list items alongside the list in `searchShareableList` query result

nina-py:
## Goal

Unblock OSL-167 (admin frontend) by adding list items to results returned by `searchShareableList`.

- Reworked the schemas to move `ShareableListItem` into the shared part and create a new interface for `ShareableList` types. The public graph implements the interface with all the required properties and the end result is `ShareableList` type, while the admin graph adds two extra properties and sets up the `ShareableListComplete` type.

- Updated fragments and sample queries/mutations to work with the changes to the schema.

- Additionally, following a Slack conversation with @jpetto and @katerinachinnappan, removed `userId` field from `ShareableList` types. 

One notable advantage of this refactor is that the Admin and Public graphs within this service don't have the same type name for shareable lists; this should make it easier to migrate to contracts in a unified admin+client graph down the track. It also aligns type names used internally (e.g. `ShareableListComplete`) with schema names.

## Tickets

- https://getpocket.atlassian.net/browse/OSL-267


jpetto (approved):
this looks great! nice use of an interface 😄

PR #108, opened 2023-03-15
Katerina/osl 265 application redis infra setup cdktf repair

cmharlow:
## Goal

update cdktf so can use latest pocket-tools/terraform-modules

### DIFF
--- a/.aws/src/main.ts
+++ b/.aws/src/main.ts
@@ -1,12 +1,12 @@
-import { Construct } from 'constructs';
-import {
-  App,
-  DataTerraformRemoteState,
-  RemoteBackend,
-  TerraformStack,
-} from 'cdktf';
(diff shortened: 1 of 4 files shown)


.aws/src/main.ts
```diff
@@ -71,8 +75,6 @@ class ShareableListsAPI extends TerraformStack {
       //So instead we set it to null and allow anything within the vpc to access it.
       //This is not ideal..
       //Ideally we need to be able to add security groups to the ALB application.
-      overrideEngineVersion: '7.0',
```
cmharlow, line 74: up to you if you want to specify this, but these values should be the tf module default now (and work as a default now).

katerinachinnappan (approved):
🚀

PR #129, opened 2023-03-24
OSL-318: resolve User obj and itemId in admin resolver

katerinachinnappan:
## Goal

Need to resolve the User obj and itemId in the admin resolver.

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-318](https://getpocket.atlassian.net/browse/OSL-318)


jpetto (approved):
nice!

PR #132, opened 2023-03-28
chore(test data): Update seed script to make lists public

nina-py:
## Goal

Improve ease of Admin Tools development for Shareable Lists.

- Update seed script to make lists public

- And have slugs generated for them so that they can be used in Dev Curation Admin Tools for testing and development

## Tickets

 - not ticketed, but related to OSL - 167

### DIFF
--- a/prisma/seed.ts
+++ b/prisma/seed.ts
@@ -39,13 +39,15 @@ async function main() {
       userId: randomUser.userId,
     });
 
-    // turn it public so that it obtains a slug - this seed data
-    // is created for the benefit of testing the admin tools after all
-    await updateShareableList(
-      prisma,
-      { externalId: list.externalId, status: ListStatus.PUBLIC },
-      randomUser.userId
-    );
+    // Turn some lists public so that the status changes and a slug is generated.
+    // This seed data is created for the benefit of testing the admin tools after all.
+    if (Math.random() > 0.5) {
+      await updateShareableList(
+        prisma,
+        { externalId: list.externalId, status: ListStatus.PUBLIC },
+        randomUser.userId
+      );
+    }
 
     // add between 5 and 10 Pocket stories to this list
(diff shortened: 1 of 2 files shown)


prisma/seed.ts
```diff
@@ -38,6 +39,14 @@ async function main() {
       userId: randomUser.userId,
     });
 
+    // turn it public so that it obtains a slug - this seed data
+    // is created for the benefit of testing the admin tools after all
+    await updateShareableList(
+      prisma,
+      { externalId: list.externalId, status: ListStatus.PUBLIC },
```
jpetto, line 46: is there value is keeping some lists private? or is this seed script only in service of the admin tool?

jpetto (approved):
just one non-blocking question 😄

katerinachinnappan (approved):

prisma/seed.ts
nina-py: I think there is, let me try and make only some lists public as I need to see how the component looks in both cases in the app itself, and not just in test coverage.

prisma/seed.ts
```diff
@@ -38,6 +39,16 @@ async function main() {
       userId: randomUser.userId,
     });
 
+    // Turn some lists public so that the status changes and a slug is generated.
+    // This seed data is created for the benefit of testing the admin tools after all.
+    if (Math.random() > 0.5) {
```
jpetto, line 44: as long as you have faith in `Math.random()` being greater than `0.5` at least once in 50 chances 😅

jpetto (approved):

prisma/seed.ts
nina-py: Well, strictly speaking, we're all right even if it's not an even split between public and private lists in seed data, we just need a bit of variety! :)

PR #133, opened 2023-03-29
feat(list items): [OSL-332] make itemId required when creating/updating list item

nina-py:
## Goal

Ensure `itemId` is always provided when adding a story to a shareable lists.

- Updated Prisma schema to make `itemId` required and added a migration (*note*: reverted this in a later commit - in the database, `itemId` will remain optional to accommodate data already in the production database).

- Updated GraphQL schema and internal types to make sure `itemId` is a required field everywhere.

## Tickets

- https://getpocket.atlassian.net/browse/OSL-332


jpetto (approved):
thanks for refactoring! ship it!

PR #138, opened 2023-03-31
chore(docs): Add instructions to reset data on Dev

nina-py:
## Goal

As above - link to the freshly created instructions page, too.

## Tickets

- unticketed


jpetto (approved):
thank you for this!

PR #139, opened 2023-03-31
chore(type safety): [OSL-346] Leverage Prisma-generated types for custom return types

nina-py:
## Goal

I stumbled across this part of Prisma documentation recently: https://www.prisma.io/docs/concepts/components/prisma-client/advanced-type-safety and realised Prisma has solved a problem that we've been solving on our own for a while by setting up custom types for return types that contain partial Prisma data  - and they always contain partial data since we don't reveal the internal integer ID, and often other table fields are also not needed to be exposed in the graph.

So I gave it a go and it works! I have had to get TypeScript to ignore a couple of lines though. It seems counter-intuitive to me that `ts-ignore` lines are needed where the docs claim things should just work. I'll dig in deeper. 

## Implementation details

- Moved away from custom types that omit the props in the data model that we don't want to expose on the graph or return from resolvers in any way (e.g., the actual internal ID of a list) and leverage Prisma-generated types and validators instead.

- Also did a brief sweep of the repo for typos and out-of-date comments.

## Tickets

- https://getpocket.atlassian.net/browse/OSL-346


src/database/types.ts
```diff
@@ -40,12 +26,81 @@ export enum ShareableListModerationReason {
...
+    description: true,
+    status: true,
+    moderationStatus: true,
+    createdAt: true,
+    updatedAt: true,
+  },
+  include: { listItems: true },
+});
+// @ts-ignore
+export type ShareableList = Prisma.PromiseReturnType<typeof shareableList>;
```
hyperparabolic, line 77: Let's try to unpack the error message that comes up when you remove the @ts-ignore here:

```
Type '{ select: { externalId: true; userId: true; slug: true; title: true; description: true; status: true; moderationStatus: true; createdAt: true; updatedAt: true; }; include: { listItems: true; }; }' does not satisfy the constraint '(...args: any) => Promise<any>'.
  Type '{ select: { externalId: true; userId: true; slug: true; title: true; description: true; status: true; moderationStatus: true; createdAt: true; updatedAt: true; }; include: { listItems: true; }; }' provides no match for the signature '(...args: any): Promise<any>'.
```

Let's just dissect how to understand this for a second without worrying about where we're going.

Quite the bit of word soup, but let's just try to find the important part to focus on. I see `provides no match for the signature '(...args: any): Promise<any>`, and that jumps out at me. The type we're feeding into the generic type variable isn't valid.

Next question: What is `Prisma.PromiseReturnType` and what does it expect?

```
type Prisma.PromiseReturnType<T extends (...args: any) => Promise<any>> = ReturnType<T> extends PromiseLike<infer U> ? U : ReturnType<T>
```

There are two parts to this:
- `T extends (...args: any) => Promise<any>` is the generic variable definition.  This tells you the type you must provide to the generic.
- `= ReturnType<T> extends PromiseLike<infer U> ? U : ReturnType<T>` is the type that the generic creates.

It is complaining because `typeof shareableList` is an object, and it wants `T extends (...args: any) => Promise<any>`.

Huh. What does something that actually satisfies this constraint look like?  I can clear the error with a line like this:

```
export type ShareableListItem = Prisma.PromiseReturnType<
  (...args: any) => Promise<typeof shareableList>
>;
```

But that doesn't really feel right... What are args even? Why does the generic make me jump through all of these hoops? Oh wait, that's actually the same type as the db query:

```
function getShareableList(
  db: PrismaClient,
  userId: number | bigint,
  externalId: string
): Promise<ShareableList>
```

When this is fed into the generic:
```
// top of the file
import { getShareableList } from './queries';
....
export type ShareableListItem = Prisma.PromiseReturnType<
  typeof getShareableList
>;
```

The original error clears, and is replaced with something else:
`Type alias 'ShareableList' circularly references itself.`

You can't magic away having to manage type definitions. Generated types are the lowest effort and closest option you're going to get.

hyperparabolic: Thought about this a bit more, and looked at their docs a bit more closely.  Let's figure out how to actually get this to work...

`Prisma.PromiseReturnType` looks like it's a tool for extracting types out of the generated client.

First in `src/database/queries/shareableList.ts` you're going to want to remove all of the awaits, and remove all of the return type annotations. Typescript instead will use the implicit response type. This is a generated prisma client type.

Let's look at the first query as an example:
```
export function getShareableList(
  db: PrismaClient,
  userId: number | bigint,
  externalId: string
) {
  // externalId is unique, but the generated type for `findUnique` here doesn't
  // include `moderationStatus`, so using `findFirst` instead
  return db.list.findFirst({
    where: {
      userId,
      externalId,
      moderationStatus: ModerationStatus.VISIBLE,
    },
    include: {
      listItems: true,
    },
  });
}
```

With the explicit response type removed, the type becomes `typeof db.list.findFirst({ ...`, or:
```
(method) Prisma.ListDelegate<Prisma.RejectOnNotFound | Prisma.RejectPerOperation>.findFirst<{
    where: {
        userId: number | bigint;
        externalId: string;
        moderationStatus: "VISIBLE";
    };
    include: {
        listItems: true;
    };
}, undefined>(args?: {
    where: {
        userId: number | bigint;
        externalId: string;
        moderationStatus: "VISIBLE";
    };
    include: {
        listItems: true;
    };
}): Prisma.Prisma__ListClient<...>
```

This looks like a big complicated mess, but we need this big complicated mess for typescript to actually understand what this function does, and what it really returns.

`Prisma.PromiseReturnType` helps you cut through the mess.  Back in the types file:
```
// top of the file
import { getShareableList } from './queries';
...
export type ShareableList = Prisma.PromiseReturnType<typeof getShareableList>;
```

This evaluates to:
```
type ShareableList = List & {
    listItems: ListItem[];
}
```

This might require some cleanup elsewhere, but this probably gets you to the end result you actually want.

Lean into the complexity and be explicit.

nina-py: Thanks Spencer! On looking at the code with fresh eyes on Monday, I realised I applied a solution from the docs to a problem we didn't even have... I couldn't get the initial solution working so I jumped to the next section of the doc without fully understanding what exactly it was solving. It did suppress most TypeScript errors but didn't actually get us the type safety I was looking for. 

I've since got the initial suggestion from the docs working - what I needed to do was tweak the data returned by some of the DB resolvers and it worked!

Herraj (approved):

katerinachinnappan (approved):
🔥

PR #140, opened 2023-04-04
OSL-371: Enfore itemId to be required at Prisma level

katerinachinnappan:
## Goal

`itemId` field has been backfilled now, so the field can be enforced to be required on the Prisma level. Cleaned up some code for passing a null `itemId`.

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-371](https://getpocket.atlassian.net/browse/OSL-371)

### DIFF
--- a/src/database/mutations/ShareableListItem.ts
+++ b/src/database/mutations/ShareableListItem.ts
@@ -61,13 +61,7 @@ export async function createShareableListItem(
 
   const input = {
     // coerce itemId to a number to conform to db schema
-    // if no itemId is set, send null. accepting null here should only
-    // occur in test scenarios, as we need to be able to mock list items that
-    // are missing itemId. as of this writing, the schema requires itemId.
-    // once we backfill old data missing itemId, we can revert this to:
-    // itemId: parseInt(data.itemId)
-    // https://getpocket.atlassian.net/browse/OSL-338
-    itemId: data.itemId ? parseInt(data.itemId) : null,
+    itemId: parseInt(data.itemId),
     url: data.url,
     title: data.title ?? undefined,
     excerpt: data.excerpt ?? undefined,
--- a/src/shared/resolvers/fields/PrismaBigInt.ts
+++ b/src/shared/resolvers/fields/PrismaBigInt.ts
@@ -1,7 +1,7 @@
 import * as Sentry from '@sentry/node';
 
 /**
- * Return a Number.
+ * Return a Number or Null (if not resolved).
  * Prisma returns a String when it retrieves the BigInt value stored in
  * the datastore (by appending a "n" char to the value).
  *
@@ -15,7 +15,7 @@ export const PrismaBigIntResolver = (parent, args, context, info) => {
   return parseFieldToInt(parent[info.fieldName]);
 };
 
-function parseFieldToInt(field: string) {
+export function parseFieldToInt(field: string): number | null {
   // as the schema now requires returning a value even when the db is missing the value
   // if !field, we should error to Sentry.
   if (!field) {
--- a/src/shared/resolvers/fields/PrismaBigInt.ts
+++ b/src/shared/resolvers/fields/PrismaBigInt.ts
@@ -16,11 +16,11 @@ export const PrismaBigIntResolver = (parent, args, context, info) => {
 };
 
 export function parseFieldToInt(field: string): number {
+  const parsedField = parseInt(field);
   // if for some reason the value is corrupt in the db, log to Sentry
-  if (!field || isNaN(parseInt(field))) {
+  if (!field || isNaN(parsedField)) {
     Sentry.captureException('Failed to parse itemId');
     return null;
   }
-  const parsedField = parseInt(field);
   return parsedField;
 }
--- a/src/shared/resolvers/fields/PrismaBigInt.ts
+++ b/src/shared/resolvers/fields/PrismaBigInt.ts
@@ -1,7 +1,7 @@
 import * as Sentry from '@sentry/node';
 
 /**
- * Return a Number or Null (if not resolved).
+ * Return a Number.
  * Prisma returns a String when it retrieves the BigInt value stored in
  * the datastore (by appending a "n" char to the value).
  *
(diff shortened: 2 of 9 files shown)


src/database/mutations/ShareableListItem.ts
```diff
@@ -61,13 +61,7 @@ export async function createShareableListItem(
...
   const input = {
     // coerce itemId to a number to conform to db schema
-    // if no itemId is set, send null. accepting null here should only
-    // occur in test scenarios, as we need to be able to mock list items that
-    // are missing itemId. as of this writing, the schema requires itemId.
-    // once we backfill old data missing itemId, we can revert this to:
-    // itemId: parseInt(data.itemId)
-    // https://getpocket.atlassian.net/browse/OSL-338
-    itemId: data.itemId ? parseInt(data.itemId) : null,
+    itemId: parseInt(data.itemId),
```
Herraj, line 64: I don't have context but is `itemId` defined as a `string` in gql schema?

src/shared/resolvers/fields/PrismaBigInt.ts
```diff
@@ -14,14 +16,12 @@ export const PrismaBigIntResolver = (parent, args, context, info) => {
...
+  // if !field, we should error to Sentry.
   if (!field) {
-    return '';
+    Sentry.captureException('itemId cannot be null');
+    return null;
   }
   const parsedField = parseInt(field);
-  // same here - if the parsed value is NaN, fall back to an empty string
-  return isNaN(parsedField) ? '' : parsedField;
+  return isNaN(parsedField) ? null : parsedField;
```
Herraj, line 26: Would be nice to add a return type of `number | null` to the function and update the function comment!

Herraj (commented):
Wondering if we should add a unit test for the `parseFieldToInt` function?

src/database/mutations/ShareableListItem.ts
katerinachinnappan: @Herraj its `ID` type which is interpreted as a string I believe, but we store it as a BigInt in the db

src/shared/resolvers/fields/PrismaBigInt.ts
jpetto: clarification: if `isNaN(parsedField)` is true, that means we have a value in the database that cannot be parsed into a number, right? if so, that would be a big problem 😅. 

in other words, this function should _never_ return anything _except_ a `number`, right?

src/shared/resolvers/fields/sharedResolversFields.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+  describe('PrismaBigInt Resolver - parseFieldToInt function', () => {
+    it('should return null if field is string of chars', async () => {
+      const itemId = 'abc';
+      const resolvedValue = parseFieldToInt(itemId);
+      expect(resolvedValue).to.be.null;
+    });
+    it('should return null if field is null', async () => {
+      const itemId = null;
+      const resolvedValue = parseFieldToInt(itemId);
+      expect(resolvedValue).to.be.null;
```
Herraj, line 25: We should also assert the sentry stub to be called once in this test case

jpetto (changes requested):
one change request on the field resolver, and i think we should update [this helper](https://github.com/Pocket/shareable-lists-api/blob/main/src/public/resolvers/utils.ts#L84) to not allow a missing/empty `itemId`.

src/shared/resolvers/fields/sharedResolversFields.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+    });
+    it('should successfully resolve value to int', async () => {
+      const itemId = '12345';
+      const resolvedValue = parseFieldToInt(itemId);
+      expect(resolvedValue).to.equal(parseInt(itemId));
```
Herraj, line 30: This is ok and works but to have a more "truer" test, we should replace the `equal(parseInt(item))` with `equal(12345)`

The reason I say this is because we want to treat our function call, `parseFieldToInt` as a black box. So whatever value it returns we assert it against what we expect. What you have now (`parseInt(itemId)`) is basically mimicking the inner workings of that function, which kind of defeats the purpose of testing. If for some reason our function logic was incorrect we'd never catch the bug in our test.

Does that make sense? 😄 once again I think it's fine in this situation since it's a very simple function and we can eyeball what `parseInt("12345")` would return.

src/shared/resolvers/fields/PrismaBigInt.ts
katerinachinnappan: @jpetto that is right, if `validateItemId` function is already validating the value prior to inserting it into the db, then we don't need the validation in the resolver function. It can simple return `parseInt(value)` and we don' need to log anything into Sentry then.

jpetto: i think it's a good idea to keep a check that `parseInt(field)` gives us a number, and a Sentry call if not. it would be an extra layer of protection/notice if somehow the input validation began failing.

Herraj (approved):

src/shared/resolvers/fields/PrismaBigInt.ts
```diff
@@ -13,15 +15,12 @@ export const PrismaBigIntResolver = (parent, args, context, info) => {
...
+export function parseFieldToInt(field: string): number {
+  // if for some reason the value is corrupt in the db, log to Sentry
+  if (!field || isNaN(parseInt(field))) {
+    Sentry.captureException('Failed to parse itemId');
+    return null;
   }
   const parsedField = parseInt(field);
-  // same here - if the parsed value is NaN, fall back to an empty string
-  return isNaN(parsedField) ? '' : parsedField;
+  return parsedField;
```
jpetto, line 25: not a blocker, but in the future could consider consolidating the `parseInt(field)` call into a variable 😄

src/shared/resolvers/fields/sharedResolversFields.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+    it('should return null if field is empty string', async () => {
+      const itemId = '';
+      const resolvedValue = parseFieldToInt(itemId);
+      expect(resolvedValue).to.be.null;
+      // Expect Sentry to get invoked and assert message
+      expect(sentryStub.callCount).to.equal(1);
+      expect(sentryStub.getCall(0).firstArg).to.equal('Failed to parse itemId');
+    });
+    it('should successfully resolve value to int', async () => {
+      const itemId = '12345';
```
jpetto, line 50: won't the actual value from the database be like `12345n`? if so, can we update the test to validate?

jpetto (changes requested):
one small ask on a test, but then good to go!

jpetto (approved):
lgtm! be sure to deploy to dev first (if you haven't already 😄)

PR #141, opened 2023-04-05
OSL-388: Add userId to ShareableList Snowplow type

katerinachinnappan:
## Goal

Sending `userId` as part of the ShareableList entity to Snowplow. As per [this request](https://pocket.slack.com/archives/CR6T8BNCT/p1680652916308989).

## Todos

- [x] test in dev

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-388](https://getpocket.atlassian.net/browse/OSL-388)

### DIFF
--- a/src/snowplow/events.spec.ts
+++ b/src/snowplow/events.spec.ts
@@ -89,6 +89,10 @@ describe('Snowplow event helpers', () => {
     expect(payload.shareableList.shareable_list_external_id).to.equal(
       shareableList.externalId
     );
+    // userId -> user_id
+    expect(payload.shareableList.user_id).to.equal(
+      parseInt(shareableList.userId as unknown as string)
+    );
     // expect slug to be undefined
     expect(payload.shareableList.slug).to.equal(undefined);
     // moderationStatus -> moderation_status
@@ -124,6 +128,10 @@ describe('Snowplow event helpers', () => {
     expect(payload.eventType).to.equal(
       EventBridgeEventType.SHAREABLE_LIST_UPDATED
     );
+    // userId -> user_id
+    expect(payload.shareableList.user_id).to.equal(
+      parseInt(shareableList.userId as unknown as string)
+    );
     // check that title was updated
     expect(payload.shareableList.title).to.equal('Updated random title');
     // check that description was updated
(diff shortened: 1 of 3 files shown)


src/snowplow/events.ts
```diff
@@ -22,8 +22,14 @@ import {
 function transformAPIShareableListToSnowplowShareableList(
   shareableList: ShareableListComplete
 ): SnowplowShareableList {
+  // userId should always be present, but if for some reason it cannot be parsed,
+  // return undefined as userId is not required in Snowplow schema.
+  const userId = isNaN(parseInt(shareableList.userId as unknown as string))
```
jpetto, line 27: we need to `parseInt` here because prisma returns a number like "123456789n", right?

something about having to check `isNaN` here feels off to me. this value will be coming from the database, which now guarantees a number value, right?

half of me wants to trust the database and prisma generated types, and the other half of me wants to keep this check, but add a Sentry call if `isNaN` returns `true`. i don't think we want to just fallback to an `undefined` value without any alerting (as this would signify a data integrity error).

i don't want to block on a philosophical discussion though - what do you think about adding a Sentry call here if `isNaN` returns `true`?

src/snowplow/events.spec.ts
```diff
@@ -89,6 +89,10 @@ describe('Snowplow event helpers', () => {
     expect(payload.shareableList.shareable_list_external_id).to.equal(
       shareableList.externalId
     );
+    // userId -> user_id
+    expect(payload.shareableList.user_id).to.equal(
+      parseInt(shareableList.userId as unknown as string)
```
jpetto, line 94: just curious - what error is thrown if you don't cast as a `string`?

jpetto (changes requested):
would like a Sentry call if we have a data integrity error around `userId`, but otherwise looking good to me!

src/snowplow/events.ts
katerinachinnappan: @jpetto I don't see why we can't add it! Definitely worth logging this if for some weird reason the userId value is corrupt.

src/snowplow/events.spec.ts
katerinachinnappan: @jpetto `Argument of type 'bigint' is not assignable to parameter of type 'string'.`. The typeof `shareableList.userId` is `bigInt` but when getting the data from the db and mapping it to the ShareableList graphQL type, the `n` get attached.

jpetto (approved):

PR #145, opened 2023-04-10
feat(listItemNoteVisibility): add field to the API

jpetto:
## Goal

add `listItemNoteVisibility` field to the API

- small refactor of some tests
- renamed `ShareableListStauts` graph enum to `ShareableListVisibility`

verified with web team that the enum rename is safe on their end (they only use the values, which aren't changing).

## Tickets

- https://getpocket.atlassian.net/browse/OSL-395

### DIFF
--- a/src/public/resolvers/mutations/ShareableList.integration.ts
+++ b/src/public/resolvers/mutations/ShareableList.integration.ts
@@ -133,11 +133,9 @@ describe('public mutations: ShareableList', () => {
       expect(result.body.errors[0].message).to.equal(ACCESS_DENIED_ERROR);
     });
 
-    it('should create a new List without ListItem', async () => {
-      const title = 'My list to share<script>alert("Hello World!")</script>';
-
+    it('should not cache the results', async () => {
       const data: CreateShareableListInput = {
-        title: title,
+        title: faker.lorem.words(5),
         description: faker.lorem.sentences(2),
       };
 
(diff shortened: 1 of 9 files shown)


src/public/resolvers/mutations/ShareableList.integration.ts
```diff
@@ -149,24 +147,68 @@ describe('public mutations: ShareableList', () => {
...
+      const result = await request(app)
+        .post(graphQLUrl)
+        .set(headers)
+        .send({
+          query: print(CREATE_SHAREABLE_LIST),
+          variables: { listData: data },
+        });
 
-      expect(list.title).to.equal(
+      expect(result.body.data.createShareableList.title).to.equal(
```
katerinachinnappan, line 171: is this what we expect? Though there was a bug filed for character encoding in HTML?

src/test/helpers/createShareableListHelper.ts
```diff
@@ -34,6 +35,7 @@ export async function createShareableListHelper(
     slug: data.slug ?? undefined,
     status: data.status ?? undefined,
     moderationStatus: data.moderationStatus ?? ModerationStatus.VISIBLE,
+    listItemNoteVisibility: data.listItemNoteVisibility ?? undefined,
```
katerinachinnappan, line 38: should `listItemNoteVisibility` be set to `PRIVATE` as it is the default instead of `undefined` if no value is passed?

katerinachinnappan (changes requested):

src/public/resolvers/mutations/ShareableList.integration.ts
```diff
@@ -336,23 +382,15 @@ describe('public mutations: ShareableList', () => {
           variables: { listData: data },
         });
 
-      // This mutation should not be cached, expect headers.cache-control = no-store
-      expect(result.headers['cache-control']).to.equal('no-store');
```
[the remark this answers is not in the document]
jpetto: i created a dedicated test for checking the `cache-control` header so we don't have to duplicate the check in other tests 😄

PR #151, opened 2023-04-13
refactor(field length): improve length checks for title & description

jpetto:
## Goal

improve length checks for title & description

- remove app code as checking now happens in API schema
- remove minimum character restriction

## Tickets

- https://getpocket.atlassian.net/browse/OSL-404

## Implementation Decisions

doesn't seem to make sense to force a user to enter 3 characters for a title or description. if they want a single character (emoji?) title or description, they can have it.


schema-public.graphql
```diff
@@ -104,8 +104,8 @@ type ShareableListPublic implements ShareableListInterface
 Input data for creating a Shareable List.
 """
 input CreateShareableListInput {
-  title: String! @constraint(minLength: 3, maxLength: 100)
-  description: String @constraint(minLength: 3, maxLength: 200)
+  title: String! @constraint(maxLength: 100)
```
katerinachinnappan, line 107: yay!

katerinachinnappan (approved):

PR #152, opened 2023-04-14
OSL-423: removing xss library, switching to using entities library for HTML encoding

katerinachinnappan:
## Goal

`xss` package doesn't seem to be consistent, so switching to `entities` package.

## Tickets

- More details here: [https://getpocket.atlassian.net/browse/OSL-423](https://getpocket.atlassian.net/browse/OSL-423)

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "aws-xray-sdk-core": "3.4.1",
     "aws-xray-sdk-express": "3.4.1",
     "cors": "2.8.5",
+    "entities": "^4.5.0",
     "express": "4.18.2",
     "express-validator": "^6.15.0",
     "graphql": "16.6.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -37,6 +37,7 @@
     "aws-xray-sdk-core": "3.4.1",
     "aws-xray-sdk-express": "3.4.1",
     "cors": "2.8.5",
+    "entities": "^4.5.0",
```
kschelonka, line 40:
```suggestion
    "entities": "4.5.0",
```
so renovate can manage

nina-py: I think Renovate bot will pin a new package automatically to a patch version when there's an update, so this could be left as-is

nina-py (approved):
Great choice

katerinachinnappan:
closing for now, keeping branch alive

PR #153, opened 2023-04-14
OSL-423: replace < and > with encoded chars

katerinachinnappan:
## Goal

As discussed, if using the `entities` library to use HTML encoding, `Katerina's List` transforms to `Katerina&apos; List` and the generated slug when the list is published is `katerinaapos-list`. We want to avoid this, so for now we are only encoding `<` and `>`.

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-423](https://getpocket.atlassian.net/browse/OSL-423)

## Implementation Decisions

Decided to write a small function to find and replace chars in a str which can be reused in other methods if needed.

### DIFF
--- a/src/public/resolvers/utils.ts
+++ b/src/public/resolvers/utils.ts
@@ -1,10 +1,12 @@
-import xss from 'xss';
-
 import { PrismaClient } from '@prisma/client';
 import { ForbiddenError, UserInputError } from '@pocket-tools/apollo-utils';
 
 import { IPublicContext } from '../context';
-import { ACCESS_DENIED_ERROR } from '../../shared/constants';
+import {
+  ACCESS_DENIED_ERROR,
+  GT_ENCODED,
+  LT_ENCODED,
+} from '../../shared/constants';
 import { isPilotUser } from '../../database/queries';
 
 /**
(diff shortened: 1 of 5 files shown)


src/public/resolvers/utils.ts
```diff
@@ -25,6 +27,23 @@ export async function executeMutation<T, U>(
...
+  str: string,
+  charsToFindArr: string[],
+  replaceWithCharsArr: string[]
+): string {
+  const transformedStr = charsToFindArr.reduce((acc, item, i) => {
+    const regex = new RegExp(item, 'g');
+    return acc.replace(regex, replaceWithCharsArr[i]);
+  }, str);
+  return transformedStr;
+}
```
Herraj, line 46: Haven't tried it out myself yet but why not use `String.replace()` instead of a reducer function 🤔

Herraj (approved):
lgtm!

src/public/resolvers/utils.ts
katerinachinnappan: no reason, I think it performs better if there is ever a case and a huge array is provided, but that will probably never happen. I can definitely simplify this

PR #154, opened 2023-04-19
feat(note): add list item note to schema

jpetto:
## Goal

add list item note to the schema, including ability to create a list item with a note. (no UX pattern for adding a note at create time, but here if we need it later.)

- should not return notes for the public shareable list query if their visibility is PRIVATE

## Tickets

- https://getpocket.atlassian.net/browse/OSL-392

## Implementation Decisions

- basic pass at hiding notes for public lists when their visibility is PRIVATE


src/database/queries/ShareableList.ts
```diff
@@ -67,6 +67,15 @@ export async function getShareableListPublic(
     throw new ForbiddenError(ACCESS_DENIED_ERROR);
   }
 
+  // remove all notes if they should be private
+  // may be able to do this logic in the query above
+  // https://getpocket.atlassian.net/browse/OSL-396
+  if (list.listItemNoteVisibility === Visibility.PRIVATE) {
+    list.listItems.forEach((li) => {
+      li.note = null;
```
katerinachinnappan, line 75: nit - maybe `delete li.note` ? so that it doesn't return as `null` to the client but rather doesn't return at all

jpetto: i _think_ the field has to return something to satisfy the graph request, even if it's `null`...

jpetto: nope - just tested and we can `delete` it!

katerinachinnappan (approved):

PR #155, opened 2023-04-19
feat(updateShareableListItem): add mutation and tests

jpetto:
## Goal

add `updateShareableListItem` mutation w/tests.

## Tickets

- https://getpocket.atlassian.net/browse/OSL-393

### DIFF
--- a/schema-public.graphql
+++ b/schema-public.graphql
@@ -151,6 +151,12 @@ input CreateShareableListItemWithList {
   sortOrder: Int!
 }
 
+input UpdateShareableListItemInput {
+  externalId: ID!
+  note: String @constraint(maxLength: 300)
+  sortOrder: Int
+}
+
 type Query {
   """
   Looks up and returns a Shareable List with a given external ID for a given user.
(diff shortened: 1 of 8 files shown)


schema-public.graphql
```diff
@@ -197,6 +203,12 @@ type Mutation {
     data: CreateShareableListItemInput!
   ): ShareableListItem
   """
+  Updates a Shareable List Item.
+  """
+  updateShareableListItem(
```
katerinachinnappan, line 208: PR looks good to me but wondering if we should wait for bulk add work and also update this mutation to accept array

katerinachinnappan (changes requested):

src/database/mutations/ShareableListItem.ts
```diff
@@ -89,6 +90,50 @@ export async function createShareableListItem(
...
+    // we need to include the list to check the user id
+    include: {
+      list: true,
+    },
+  });
+
+  // if no list item was found, or this list item is owned by another user,
+  // return a not found error
+  if (!listItem || parseInt(listItem.list.userId as any) !== userId) {
+    throw new NotFoundError(`A list item by that ID could not be found`);
```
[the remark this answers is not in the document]
jpetto: 404 is the safer option, as it doesn't reveal any information about information owned by other users. 403 would let an attacker know that whatever data they have stumbled upon is in fact real data.

schema-public.graphql
jpetto: decided in our review meeting that we can ship this now to unblock notes and we'll revisit the bulk update stuff in a later ticket.

katerinachinnappan (approved):

PR #156, opened 2023-04-19
OSL-157: Add SQSLambda AWS TF Infra

katerinachinnappan:
## Goal

This PR adds the AWS infra for SQSLambda and the relevant configs

## Todos

- [x] test in dev

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-157](https://getpocket.atlassian.net/browse/OSL-157)


.aws/src/SQSLambda.ts
```diff
@@ -0,0 +1,62 @@
...
+    vpc: PocketVPC,
+    region: DataAwsRegion,
+    caller: DataAwsCallerIdentity,
+    pagerDuty?: PocketPagerDuty
+  ) {
+    super(scope, name);
+
+    this.lambda = new PocketSQSWithLambdaTarget(this, 'sqs-event-consumer', {
+      name: `${config.prefix}-Sqs-Event-Consumer`,
+      batchSize: 10, // The number of records to send to the function in each batch.
```
jpetto, line 28: suggestion - it might make the lambda simpler to restrict this to 1 record per batch. we won't be getting massive amounts of user account deletions 😄. (however, if there's prior art to handling batches in a lambda, i'd actually recommend persisting that pattern.)

.aws/src/config/index.ts
```diff
@@ -37,9 +37,15 @@ export const config = {
   rds,
   cacheNodes,
   cacheSize,
+  reservedConcurrencyLimit: 1, // A maximum of 1 instance of the Lambda shall be running at whatever time needed.
```
jpetto, line 40: curious what the logic behind this is. do we always need a lambda running?

jpetto (approved):
a couple small, non-blocking questions. so close!

.aws/src/SQSLambda.ts
katerinachinnappan: I was thinking that , but yes, there is a pattern established with the configs so I went with that. I'll merge for now but always happy to open another PR if needed !

PR #159, opened 2023-04-20
CHORE: Removig caching from CircleCI and updating to cimg/node:18.16

katerinachinnappan:
## Goal

Removing `restore_cache` & `save_cache` steps from CircleCI config & updating to use `cimg/node:18.16`


hyperparabolic (approved):
👍

PR #175, opened 2023-04-26
OSL-449: SQS Lambda logic & deploy via CircleCI

katerinachinnappan:
## Goal

* Adding logic to AWS Lambda to trigger `deleteUserData` express endpoint as part of account deletion effort.
* Deploying lambda via CircleCI

## Todos

- [x] test in dev

## Tickets

- https://getpocket.atlassian.net/browse/OSL-449


lambda/events/handlers/accountDelete.ts
```diff
@@ -0,0 +1,29 @@
...
+ * to delete user data
+ * @param record SQSRecord containing forwarded event from eventbridge
+ * @throws Error if response is not ok
+ */
+export async function accountDeleteHandler(record: SQSRecord) {
+  const message = JSON.parse(JSON.parse(record.body).Message)['detail'];
+  const postBody = {
+    userId: message['userId'],
+  };
+  const res = await fetch(config.apiEndpoint + config.deleteUserDataPath, {
```
sri-kirsh, line 16: we usually have a backoff/retry for inter service background communication, incase the call fails:
e.g https://github.com/Pocket/account-data-deleter/blob/main/lambda/events/handlers/accountDelete/postRequest.ts#L23

lambda/events/index.ts
```diff
@@ -0,0 +1,32 @@
...
+        );
+      }
+      await handlers[message['detail-type']](record);
+    } catch (error) {
+      console.log(error);
+      Sentry.captureException(error);
+      batchFailures.push({ itemIdentifier: record.messageId });
+    }
+  }
+  return { batchItemFailures: batchFailures };
```
sri-kirsh, line 29: and i think we should enable this config in lambda setup if we want to report message failure within a single batch: https://github.com/Pocket/list-api/blob/main/.aws/src/lambda/base/SqsLambda.ts#L34

.circleci/continue-config.yml
```diff
@@ -0,0 +1,192 @@
+version: 2.1
+
+setup: false
```
sri-kirsh, line 3: circle looks ok at high level, dev deployment and test would validate this - so not reviewing in detail

efixler: 😅

lambda/events/index.ts
katerinachinnappan: oh, it is already enabled on line 30!

katerinachinnappan:
tested in dev: 

https://s3.console.aws.amazon.com/s3/buckets/pocket-shareablelistsapi-dev-sqs-event-consumer?region=us-east-1&tab=objects

https://us-east-1.console.aws.amazon.com/lambda/home?region=us-east-1#/functions/ShareableListsApi-Dev-Sqs-Event-Consumer-Function?tab=code

sri-kirsh (approved):

PR #180, opened 2023-04-27
Infra 1189/add basic logger

cmharlow:
## Goal

Adds a basic logger. This will be refined and moved to a package (for easier sharing across other services, not specific to apollo) after a few more iterations. Updated code & XRay to use this logger. Adds some basic http request logging with userid (if present) embedded. Creates a simple (if obtrusive) id generator per thread to embed a consistent id in logs for coordination. This should be replaced shortly with an external id (to coordinate with other systems).

Also fixes some apollo plugins issues where dev environment is not sending operations traces to apollo.

Note: this is second in a chain of PRs, so just diffing them from now on the PR ahead of them. Rebasing etc. will be done before merges to main. Already rebased and tested on dev.

## Tickets

- Part of https://getpocket.atlassian.net/browse/INFRA-1189


src/logger.ts
```diff
@@ -0,0 +1,80 @@
+// to be made into a package for easier sharing shortly
+import rTracer from 'cls-rtracer';
+import winston from 'winston';
```
efixler, line 3: Why `winston` in particular?

src/logger.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const levels = {
+  error: 0,
+  warn: 1,
+  info: 2,
+  http: 3,
+  graphql: 4,
```
bassrock, line 10: Curious why we use a level for graphql/http instead of a category?

efixler:
Super nice to see logging coming to Pocket :)

src/logger.ts
cmharlow: @efixler sorry, you weren't tagged in review - i'm still working on CI and coordinating previous work / prs with shareable lists owners.

cmharlow:
> Super nice to see logging coming to Pocket :)

It was always part of the monitoring & observability project, starting last December. I know you wrote a much smaller, less detailed proposal later on though that this work already covers. So everyone wins?

src/logger.ts
cmharlow: its part of work coming!!!!!

bassrock: yay!

efixler:
@cmharlow yes I did write a document advocating for prioritizing logging, so, yes, glad to see we're on the same page :)

PR #182, opened 2023-04-28
OSL-472: adding listRestorationReason field to db, updating resolvers and tests

katerinachinnappan:
## Goal

Adding a new optional field to the List obj `listRestorationReason` to contain the reason why a list was restored by a moderator (moderation status from hidden to visible). See [slack thread](https://pocket.slack.com/archives/C04HWMUKLN5/p1682626870569429) here.

* Added optional field to prisma schema.
* Added the field as an optional param to the admin graphQL schema.
* `moderationReason` is now optional on the graph level for the `ModerateShareableInput` type but enforced in the resolver when list is set to hidden.
* `listRestoration` is also optional on the graph level for the `ModerateShareableInput` type but enforced in the resolver when list is set to visible.
* Updated tests.

For now, we will not store historical moderation data in the db, and overwrite the moderation fields if a single list is being moderated back and forth.

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-472](https://getpocket.atlassian.net/browse/OSL-472)

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -34,7 +34,7 @@ model List {
   moderatedBy             String?          @db.VarChar(255)
   moderationReason        String?          @db.Text
   moderationDetails       String?          @db.Text
-  listRestorationReason   String?          @db.Text
+  restorationReason       String?          @db.Text
   listItemNoteVisibility  Visibility       @default(PRIVATE)
   createdAt               DateTime         @default(now()) @db.DateTime(0)
   updatedAt               DateTime         @updatedAt
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -34,6 +34,7 @@ model List {
   moderatedBy             String?          @db.VarChar(255)
   moderationReason        String?          @db.Text
   moderationDetails       String?          @db.Text
+  listRestorationReason   String?          @db.Text
   listItemNoteVisibility  Visibility       @default(PRIVATE)
   createdAt               DateTime         @default(now()) @db.DateTime(0)
   updatedAt               DateTime         @updatedAt
--- a/src/admin/resolvers/mutations/ShareableList.ts
+++ b/src/admin/resolvers/mutations/ShareableList.ts
@@ -36,16 +36,15 @@ export async function moderateShareableList(
       throw new UserInputError(`Moderation reason required.`);
     }
   }
-  // if list is set to visible, enforce listRestorationReason is also passed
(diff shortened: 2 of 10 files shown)


efixler (approved):
It looks like we're eventually driving towards having a list of moderation notes, but this is sensible and looks good!

src/admin/resolvers/mutations/ShareableList.ts
```diff
@@ -29,5 +30,22 @@ export async function moderateShareableList(
...
+    if (!data.moderationReason) {
+      throw new UserInputError(`Moderation reason required.`);
+    }
+  }
+  // if list is set to visible, enforce listRestorationReason is also passed
+  if (data.moderationStatus === ModerationStatus.VISIBLE) {
+    if (
+      !data.listRestorationReason ||
+      (data.listRestorationReason &&
+        data.listRestorationReason.trim().length === 0)
```
jpetto, line 44: just verifying - the reason we're checking the length with `listRestorationReason` and not with `moderationReason` is because the former is free text and the latter from a drop down/enum, is that right?

src/database/mutations/ShareableList.ts
```diff
@@ -212,6 +212,7 @@ export async function moderateShareableList(
       isShareableListEventType: true,
     });
   }
+  // TODO: send shareable-list-visible event to Snowplow - OSL-473
```
jpetto, line 215: love comments that reference tickets!

prisma/schema.prisma
```diff
@@ -34,6 +34,7 @@ model List {
   moderatedBy             String?          @db.VarChar(255)
   moderationReason        String?          @db.Text
   moderationDetails       String?          @db.Text
+  listRestorationReason   String?          @db.Text
```
jpetto, line 37: my only change request is to be consistent on naming and make this `restorationReason` to match `moderationReason`.

jpetto (changes requested):
one small change request, but otherwise this is great in terms of code and scope.

src/admin/resolvers/mutations/ShareableList.ts
katerinachinnappan: @jpetto that is correct!

prisma/schema.prisma
katerinachinnappan: @jpetto totally makes sense, will do

jpetto (approved):
🚢

PR #191, opened 2023-05-03
OSL-478: Adding updateShareableListItems mutation for bulk edit feat

katerinachinnappan:
## Goal

Adding a new `updateShareableListItems` mutation which accepts an array of shareable list item `sortOrder`. This is for bulk edit/ reorder effort.

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-478](https://getpocket.atlassian.net/browse/OSL-478)

## Implementation Decisions

According to this [known issue](https://github.com/prisma/prisma/issues/6862), it is not possible to use the `updateMany` prisma call to update several rows with different values, so using the [transactions API ](https://www.prisma.io/docs/guides/performance-and-optimization/prisma-client-transactions-guide#transaction-api)to perform the bulk update.

### DIFF
--- a/schema-public.graphql
+++ b/schema-public.graphql
@@ -151,10 +151,21 @@ input CreateShareableListItemWithList {
   sortOrder: Int!
 }
 
+"""
+Input data for updating a single Shareable List Item.
+"""
 input UpdateShareableListItemInput {
   externalId: ID!
   note: String @constraint(maxLength: 300)
   sortOrder: Int
+}
+
+"""
+Input data for updating an array of Shareable List Items, targeting sortOrder.
+"""
+input UpdateShareableListItemsInput {
+  externalId: ID!
+  sortOrder: Int!
 }
 
 type Query {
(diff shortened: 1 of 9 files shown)


schema-public.graphql
```diff
@@ -203,12 +214,18 @@ type Mutation {
...
+  Updates a single Shareable List Item.
   """
   updateShareableListItem(
     data: UpdateShareableListItemInput!
   ): ShareableListItem!
   """
+  Updates an array of Shareable List Items (sortOrder).
+  """
+  updateShareableListItems(
+    data: [UpdateShareableListItemsInput!]!
```
kschelonka, line 226: I'd advocate a limit here so that no one can take down the db with a massive transaction can use `@constraint(minItems: 1, maxItems: XX)`

jpetto: agree with @kschelonka here - we need to have some kind of limit to protect ourselves. i brought this up recently (when? where? idk 🙃 ), but @anthony-liddle mentioned the web team already limits these calls in batches of 30, which i think is a very reasonable limit. i'd be fine starting with a max of 30 and increasing from there if/when the need arises.

src/public/resolvers/mutations/ShareableListItem.integration.ts
```diff
@@ -718,6 +720,224 @@ describe('public mutations: ShareableListItem', () => {
...
+        .set(headers)
+        .send({
+          query: print(UPDATE_SHAREABLE_LIST_ITEMS),
+          variables: { data },
+        });
+
+      expect(result.body.errors[0].extensions.code).to.equal('NOT_FOUND');
+      expect(result.body.errors[0].message).to.contain(
+        'Error - Not Found: A list item by that ID could not be found'
+      );
```
jpetto, line 883: i generally do not advocate to test the behavior of a library, BUT since transactions are a preview feature, i think it would be nice (and make me feel more confident) if we added a check in this test that the valid list items have _not_ been updated. (could just do a direct prisma query and validate the returned data.) too many years in software has made me suspicious 😅

jpetto (changes requested):
this is great work. agree we need a max limit, and one request for an additional test (only bc transactions are a preview feature).

jpetto (approved):
nice! ship it!

PR #195, opened 2023-05-09
refactor(pilot access): pilot access only restricts publishing lists

jpetto:
## Goal

pilot access only restricts publishing lists.

## Tickets

- https://getpocket.atlassian.net/browse/OSL-462

## Implementation Decisions

tried to get the tests only referencing pilot users where it should. there's better refactoring to do here (in the tests), but i didn't want to block getting this out for aly's testing.


katerinachinnappan (approved):
Lgtm, but let's make a ticket / note to refactor / update tests

PR #196, opened 2023-05-10
feat(parser Item): reference parser Item from shareable list item

jpetto:
## Goal

reference parser Item from shareable list item.

## Tickets

- https://getpocket.atlassian.net/browse/OSL-482

## Implementation Decisions

not extending `Item` yet, but we will soon (adding in `ShareableListItem` from the `Item` side).

joining on `itemId`, but we could also use `url`. is there a preference at this point?

### DIFF
--- a/schema-shared.graphql
+++ b/schema-shared.graphql
@@ -59,6 +59,13 @@ type User @key(fields: "id", resolvable: false) {
 }
 
 """
+Resolve by reference the Item entity to connect ShareableListItems to Items.
+"""
+type Item @key(fields: "itemId", resolvable: false) {
+  itemId: String!
+}
+
+"""
 A Pocket Save (story) that has been added to a Shareable List.
 """
 type ShareableListItem {
@@ -70,6 +77,10 @@ type ShareableListItem {
   The Parser Item ID.
   """
   itemId: ID!
+  """
+  The Parser Item
+  """
+  item: Item! @cacheControl(maxAge: 60, scope: PUBLIC)
   """
   The URL of the story saved to a list.
   """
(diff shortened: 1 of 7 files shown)


schema-shared.graphql
```diff
@@ -58,6 +58,13 @@ type User @key(fields: "id", resolvable: false) {
   id: ID!
 }
 
+"""
+Resolve by reference the Item entity to connect ShareableListItems to Items.
+"""
+type Item @key(fields: "itemId", resolvable: false) {
+  itemId: String!
+}
+
```
Herraj, line 67: I'm confused as to why we are adding the `resolvable: false`) in `@key` directive and then defining a resolver for `Item` type 🤔

My understanding is that adding `resolvable: false` means that this subgraph can't resolve it and Parse/parser-graphql-wrapper subgraph should resolve it?

kschelonka: Yes, it allows referencing an entity without contributing any additional fields

https://www.apollographql.com/docs/federation/entities/#referencing-an-entity-without-contributing-fields

kschelonka (approved):

schema-shared.graphql
Herraj: Yes, but what is the resolver for? Shouldn't the other subgraph be resolving the `Item.itemId` for now? At least, until we add more `Item` fields in this subgraph?

jpetto: @Herraj you are correct. i had to add the field level resolver to populate the `itemId` key of the shared type (`Item`). otherwise (i think!) there's no way for apollo to automatically know to use the `itemId` property of the `ShareableListItem` type as that `key` value.

PR #198, opened 2023-05-11
Infra 1169/synthetic skeleton

cmharlow:
## Goal

Synthetic Alerts skeleton. Has some sane defaults.

Sets up synthetic alert (in vpc) for uptime checks (e.g. http endpoint on get returns 2xx) & "query" checks (e.g. http post on graphql endpoint with graphql sent & jmespath + expected response to validate response).

Removes old alerts from setup (default being load balance 5xx checks, which have failed us in the past).

Long term, the idea is to have this synthetic setup as generic as possible for uptime & graphql checks, then spin up needed synthetics via that config map structure (which just maps to environment variables for the synthetic checks).

## Todos

Shouldn't block this PR going through, but immediate next steps

- [ ] Get @wtfluckey & @jpetto plus others in shareable lists work to add to / validate endpoints we should check
- [ ] link these checks (once in dev & prod) to shareable lists service doc & SLIs where relevant
- [ ] move the synthetic check javascript code to its own repo so it can be deployed to s3 & reused across many synthetic check instantiations (and avoid the insanity of cdktf ts > tf json > tf resource requiring js > ts/js > zip bundling process
- [ ] move the synthetic check terraform to a module for ease of reuse
- [ ] update alerting docs with examples & further details on using this setup

### DIFF
--- a/.aws/src/main.ts
+++ b/.aws/src/main.ts
@@ -27,26 +27,27 @@ import {
 } from 'cdktf';
 import * as fs from 'fs';
 import { SQSLambda } from './SQSLambda';
+import { APIMonitoring } from './synthetics';
 
 class ShareableListsAPI extends TerraformStack {
   constructor(scope: Construct, name: string) {
     super(scope, name);
 
+    new ArchiveProvider(this, 'archive-provider');
     new AwsProvider(this, 'aws', { region: 'us-east-1' });
     new LocalProvider(this, 'local_provider');
-    new ArchiveProvider(this, 'archive-provider');
     new NullProvider(this, 'null_provider');
     new PagerdutyProvider(this, 'pagerduty_provider', { token: undefined });
-
     new RemoteBackend(this, {
       hostname: 'app.terraform.io',
       organization: 'Pocket',
       workspaces: [{ prefix: `${config.name}-` }],
     });
 
+    const caller = new DataAwsCallerIdentity(this, 'caller');
     const pocketVpc = new PocketVPC(this, 'pocket-vpc');
     const region = new DataAwsRegion(this, 'region');
-    const caller = new DataAwsCallerIdentity(this, 'caller');
+
     const cache = ShareableListsAPI.createElasticache(this, pocketVpc);
     const sqsLambda = new SQSLambda(
       this,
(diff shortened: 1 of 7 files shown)


.aws/src/main.ts
```diff
@@ -80,6 +83,38 @@ class ShareableListsAPI extends TerraformStack {
...
+        ],
+        securityGroupIds: pocketVpc.defaultSecurityGroups.ids,
+        subnetIds: pocketVpc.privateSubnetIds,
+        uptime: [
+          {
+            response: 'ok',
+            url:
+              config.environment === 'Prod'
+                ? 'shareablelistsapi.readitlater.com/.well-known/apollo/server-health'
+                : 'shareablelistsapi.getpocket.dev/.well-known/apollo/server-health',
```
jpetto, line 110: here too?

.aws/src/synthetics.ts
```diff
@@ -0,0 +1,272 @@
...
+        startCanary: true,
+        vpcConfig: {
+          subnetIds: checkConfigs.subnetIds,
+          securityGroupIds: checkConfigs.securityGroupIds,
+        },
+        zipFile: syntheticZipFile.outputPath,
+      });
+
+      new CloudwatchMetricAlarm(this, 'synthetic_check_alarm_uptime', {
+        alarmDescription: `Alert when ${check.name} canary success percentage has decreased below 66% in the last 15 minutes`,
```
jpetto, line 202: what's the reasoning behind this threshold? it sounds reasonable, just curious.

jpetto (approved):
no blockers, though you can maybe use existing config values for the dev/prod domains.

this looks like a very solid pattern for monitoring - excited to see if evolve and get abstracted.

.aws/src/main.ts
```diff
@@ -80,6 +83,38 @@ class ShareableListsAPI extends TerraformStack {
...
+
+    const monitoring = new APIMonitoring(this, 'monitoring');
+    monitoring.createSyntheticChecks(
+      {
+        query: [
+          {
+            endpoint:
+              config.environment === 'Prod'
+                ? 'shareablelistsapi.readitlater.com'
+                : 'shareablelistsapi.getpocket.dev',
```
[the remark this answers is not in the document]
cmharlow: good callout. I meant to return to variable-ize away these fields, thanks for catching that I didn't do that. updating now.

.aws/src/synthetics.ts
cmharlow: Honestly, just a place to start. I'm happy to adjust. Figure 15 minutes is long enough that its probably not self-heal-able, but that's just based off of my own experiences and not any real data.

.aws/src/main.ts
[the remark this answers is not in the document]
cmharlow: done

.aws/src/synthetics.ts
cmharlow: leaving open for further discussion.

PR #200, opened 2023-05-12
chore(shareableListPublic): add listItemNoteVisibility to query

jpetto:
## Goal

add listItemNoteVisibility to `shareableListPublic` query

## Tickets

- https://getpocket.atlassian.net/browse/OSL-503

## Implementation Decisions

even though they are now equal, kept the `ShareableListPublicProps` and `ShareableListPublicQueryProps` fragments separate, as these may well diverge in the near future.

### DIFF
--- a/src/public/resolvers/queries/ShareableList.integration.ts
+++ b/src/public/resolvers/queries/ShareableList.integration.ts
@@ -530,7 +530,7 @@ describe('public queries: ShareableList', () => {
         2
       );
 
-      // default PRIVATE listItemNoteVisibility
+      // listItemNoteVisibility is PUBLIC in this scenario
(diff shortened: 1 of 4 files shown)


src/public/resolvers/queries/ShareableList.integration.ts
```diff
@@ -527,6 +530,11 @@ describe('public queries: ShareableList', () => {
         2
       );
 
+      // default PRIVATE listItemNoteVisibility
```
katerinachinnappan, line 533: Nit pick - this comment shouldn't be here ? List item note viability is public in this scenario

katerinachinnappan (approved):

PR #205, opened 2023-05-16
OSL-510: adding an admin shareable list item type

katerinachinnappan:
## Goal

**Problem:** Curation admin tool (or any client using the admin graph) generates the admin types and uses them to get data. Parser `Item` type generated by the admin tool has `itemId`, `givenUrl` and `normalUrl` required. The shareable lists API only resolves `itemId` on `Item`. Front-end tool complains about type mismatch as it uses the generated types. 

**Solution:** The admin graph does not need `item: Item` returned for shareable list item, so creating an admin type to use specifically for the admin graph. 

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-510](https://getpocket.atlassian.net/browse/OSL-510)

### DIFF
--- a/schema-admin.graphql
+++ b/schema-admin.graphql
@@ -63,7 +63,7 @@ type ShareableListComplete implements ShareableListInterface {
   """
   Pocket Saves that have been added to this list by the Pocket user.
   """
-  listItems: [ShareableListItem!]!
+  listItems: [ShareableListItemAdmin!]!
   """
   The visibility of notes added to list items for this list.
   """
--- a/schema-admin.graphql
+++ b/schema-admin.graphql
@@ -21,7 +21,65 @@ enum ShareableListModerationReason {
   PHISHING
 }
 
-type ShareableListComplete implements ShareableListInterface {
+"""
+A Pocket Save (story) that has been added to a Shareable List. This is the admin version which
+does not include the Parser Item.
+"""
+type ShareableListItemAdmin {
+  """
+  A unique string identifier in UUID format.
+  """
+  externalId: ID!
+  """
+  The Parser Item ID.
+  """
+  itemId: ID!
+  """
+  The URL of the story saved to a list.
+  """
+  url: Url!
+  """
+  The title of the story. Supplied by the Parser.
+  May not be available for URLs that cannot be resolved.
+  Not editable by the Pocket user, as are all the other
+  Parser-supplied story properties below.
+  """
+  title: String
+  """
+  The excerpt of the story. Supplied by the Parser.
+  """
+  excerpt: String
+  """
+  User generated note to accompany this list item.
+  """
+  note: String
+  """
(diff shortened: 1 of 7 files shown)


schema-shared.graphql
```diff
@@ -22,6 +22,11 @@ A URL to a web page or image.
 """
 scalar Url
 
+"""
+An abstract GraphQL type which returns a public or admin ShareableListItem.
+"""
+union ShareableListItemUnion = ShareableListItem | ShareableListItemAdmin
+
```
nina-py, line 29: If it's only for the Admin graph, should it be in both the Public and Admin graphs? Is there another way to do it without adding this to Client API?

efixler: I'm inferring that the intent here is to minimize divergence between the admin and public versions of the same thing and the code bits that handle them, since they're nearly synonymous.  I can see how there's tradeoffs in both directions here, either using a union/common interface, or more sharply separating the two.

I'm wondering why you decided to not just go in the other direction and resolve the two urls fields that admin is expecting. Is this just to conserve data size, or is there another reason? It might just simplify things to be type uniform everywhere (granting that there may be other considerations in play)

jpetto (changes requested):
proposed a small and medium size suggestion on how to more cleanly separate the schemas. apologies for blowing up the scope of this a bit, but i think it'll get us to a better place. 😄

schema-shared.graphql
jpetto: @efixler i think we don't want to resolve those url fields for a confluence reasons - but mostly because we don't want shareable lists API to be in the service of resolving data it doesn't own. we _could_ resolve `givenUrl` just because we happen to store it, but we couldn't do the same for `normalUrl` (there's some logic behind resolving that field).

i think another factor is the issue we're trying to work around - that being the admin tool's generated types. i personally don't love adding code to satisfy that kind of problem - particularly when we need to create a custom type for the admin tool anyway. different clients, different types. (also why i don't love code generators, but in this case they do make a lot things easier.)

schema-admin.graphql
```diff
@@ -21,7 +21,65 @@ enum ShareableListModerationReason {
...
+"""
+type ShareableListItemAdmin {
+  """
+  A unique string identifier in UUID format.
+  """
+  externalId: ID!
+  """
+  The Parser Item ID.
+  """
+  itemId: ID!
```
jpetto, line 36: could you still add `item: Item` (not required) here? or does that still break generated types? might be nice to be able to still request the parser metadata through that connection.

OR were you planning to do that once we drop the metadata from this service?

jpetto (approved):
one non-blocking question. nice work!

schema-admin.graphql
katerinachinnappan: @jpetto was planning on doing that after you wrap up your draft PR removing the metadata from the schema!

PR #209, opened 2023-05-18
OSL-499: Metadata refactoring - extend Parser Item type in the shareable lists API

katerinachinnappan:
## Goal

* Extending the Parser `Item` type
* Adding a `dataLoader` for `ShareableListItem`
* Adding & testing `__resolveReference` for `ShareableListItem`

## Todos

- [x] push to dev

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-499](https://getpocket.atlassian.net/browse/OSL-499)


src/dataLoaders/shareableListItemLoader.ts
```diff
@@ -0,0 +1,61 @@
...
+  });
+}
+
+export const sortShareableListItemsByGivenItemIds = (
+  itemIds: bigint[],
+  listItems: ShareableListItem[]
+): ShareableListItem[] => {
+  // create a map of itemIds to shareable list items
+  const itemIdsToShareableListItem = listItems.reduce((acc, listItem) => {
+    if (listItem) {
```
jpetto, line 27: do we need this check? could a `listItem` in this iterator ever be falsy?

jpetto (changes requested):
a few small questions/suggestions as-is, but the big piece piece missing is limiting results based on `userId`. at this point, the use case is solely to tie saves to shareable list items, meaning the results need to be filtered by the current user.

i think we can do as you suggested and pull the `userId` from the headers as we do for other queries. the tricky part/big question is - will headers be sent down to other subgraphs/subgraph queries? or will headers get "trapped" in the originating subgraph (parser-graphql-wrapper, in this case)? if this _doesn't_ work, we might need to ticket a blocking ticket to figure out that flow.

katerinachinnappan:
@jpetto correct me if I am wrong but does list api get the userId for saves from the headers as well -- passed via`parser-graphql-wrapper` subgraph?

katerinachinnappan:
@jpetto i think we might need to do something like this?

https://github.com/Pocket/list-api/blob/main/schema.graphql#L558

jpetto:
> @jpetto i think we might need to do something like this?
> 
> https://github.com/Pocket/list-api/blob/main/schema.graphql#L558

yes. and, what do you mean? 😆

the yes is - using that example for our case, `User` would be `Item`, and `SavedItems` would be `ShareableListItems`. i poked into the resolvers in that repo, and see that it _is_ getting `userId` from the context - which means we can hopefully do the same.

src/public/resolvers/index.ts
```diff
@@ -28,7 +31,12 @@ export const resolvers = {
   ShareableListPublic: {
     user: UserResolver,
   },
+  Item: {
+    shareableListItem,
+  },
   ShareableListItem: {
+    // resolve shareable list item by itemId Parser Item uses to reference it
+    __resolveReference: resolveShareableListItemParserItemReference,
```
kschelonka, line 39: I don't think you need this, unless I'm mistaken about what exists in other subgraphs. can other subgraphs send representations of a shareable list item?

src/public/resolvers/item.ts
```diff
@@ -0,0 +1,31 @@
...
+      // actually a number in the db (legacy problems)
+      validateItemId(itemId);
+      return await dataLoaders.shareableListItemLoader.load(parseInt(itemId));
+    } else {
+      return null;
+    }
+  } catch (err) {
+    console.log(err);
+    Sentry.captureException(err);
+    throw err;
```
kschelonka, line 29: You're double-reporting here, can remove this since you use the Sentry plugin.

src/public/resolvers/index.ts
jpetto: @kschelonka the use case is - when clients request an `Item` as a subquery of a `SavedItem`, they want to see what shareable lists that `Item` may be a part of. so, i think we do need this?

jpetto: actually wait - i think kat is right 😅. the `Item: { shareableListItem }` resolver above handles all current use-cases. (e.g. all these tests pass if you remove the `__resolveReference` line.)

the difference between this subgraph and collections (which i believe this PR was modeled on) is that the parser-graphql-wrapper extends the `Collection` type with additional fields, which (i think?) is what necesitates a `__referenceResolver` in collection-api.

we should be able to remove this resolver.

jpetto (changes requested):
re-read / re-thought kat's comment and saw her point - we should be able to drop the `__referenceResolver` here.

src/public/resolvers/index.ts
katerinachinnappan: Ahh yes that makes sense. Parser-graphql-wrapper will not add any additional properties to the shareable list item type

jpetto: now that we've switched to `SavedItem`, i think we'll still have to define that as a property in list API, which will require a `type ShareableListItem` in the schema. but it'll just be a reference (like we have for `Item` in this repo).

katerinachinnappan: yep, I have a PR coming up in a lil bit

katerinachinnappan: @jpetto I'm re-reading apollo docs and not following anymore why we still need to reference shareable list item in list API? we have already extended `SavedItem` in this repo and added an additional `shareableListItem` property which we also resolve in this repo. 

So a user can check if a save item is also a shareable list item by drilling into the user -> save item -> shareable list item.

What do we achieve with referencing `ShareableListItem` type in list API? we aren't adding any additional properties to `ShareableListItem` type in List API.

jpetto: you're totally right 😄. the graph does it for us since we're `extend`ing the type.

jpetto (approved):
this is looking good! i think we can merge and wait for the web team to test some queries.

PR #215, opened 2023-05-23
OSL-499: Extending SavedItem & adding shareableListTotalCount property

katerinachinnappan:
## Goal

* Extending `SavedItem` type & adding an additional `shareableListTotalCount` property to indicate in how many shareable lists a save is (via `url`). See [slack thread](https://pocket.slack.com/archives/C04HWMUKLN5/p1684804833654229) for discussion.

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-499](https://getpocket.atlassian.net/browse/OSL-499)

### DIFF
--- a/src/shared/resolvers/fields/SavedItem.ts
+++ b/src/shared/resolvers/fields/SavedItem.ts
@@ -11,13 +11,14 @@ export async function getShareableListTotalCount(
   db: PrismaClient,
   userId: number | bigint,
   url: string
-): Promise<number> {
+): Promise<number | null> {
   if (!userId) {
     return null;
   }
-  // find shareable list items where url matches given
-  // and where parent list belongs to current user
-  const listItems = await db.listItem.findMany({
+
+  // get the total count of listItems where url is matching
+  // AND parent list belongs to current userId
+  const totalCount = await db.listItem.count({
     where: {
       url,
       list: {
@@ -25,11 +26,14 @@ export async function getShareableListTotalCount(
       },
     },
   });
-  const totalCount = listItems.length;
 
   return totalCount;
 }
 
-export async function SavedItemResolver(parent, _, { userId, db }) {
+export async function SavedItemResolver(
+  parent,
+  _,
+  { userId, db }
+): Promise<number | null> {
   return await getShareableListTotalCount(db, userId, parent.url);
 }
--- a/src/shared/resolvers/fields/SavedItem.ts
+++ b/src/shared/resolvers/fields/SavedItem.ts
@@ -1,0 +1,35 @@
+import { PrismaClient } from '@prisma/client';
+
+/**
+ * Returns the total count of shareable lists a save is in by the save url.
+ *
+ * @param db
+ * @param userId
+ * @param url
+ */
+export async function getShareableListTotalCount(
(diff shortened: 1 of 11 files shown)


src/shared/resolvers/fields/SavedItem.ts
```diff
@@ -0,0 +1,35 @@
...
+  // find shareable list items where url matches given
+  // and where parent list belongs to current user
+  const listItems = await db.listItem.findMany({
+    where: {
+      url,
+      list: {
+        userId,
+      },
+    },
+  });
```
jpetto, line 27: i wonder if we can use [prisma's filtered count syntax](https://www.prisma.io/docs/concepts/components/prisma-client/aggregation-grouping-summarizing#filtered-count) here? would be a more performant approach if it works. (apologies if you already tried it 😄.)

src/shared/resolvers/fields/SavedItem.ts
```diff
@@ -0,0 +1,35 @@
...
+  const totalCount = listItems.length;
+
+  return totalCount;
+}
+
+export async function SavedItemResolver(parent, _, { userId, db }) {
```
jpetto, line 33: could we type the return value on this function?

src/shared/resolvers/fields/SavedItem.ts
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @param db
+ * @param userId
+ * @param url
+ */
+export async function getShareableListTotalCount(
+  db: PrismaClient,
+  userId: number | bigint,
+  url: string
+): Promise<number> {
```
jpetto, line 14: this could be a `number` or `null`, right?

src/test/helpers/createShareableListItemHelper.ts
```diff
@@ -28,15 +28,15 @@ export async function createShareableListItemHelper(
...
-    itemId: data.itemId ?? faker.datatype.number(),
+    itemId: data.itemId ?? faker.number.bigInt(),
     url: data.url ?? `${faker.internet.url()}/${faker.lorem.slug(5)}`,
-    title: data.title ?? faker.random.words(5),
+    title: data.title ?? faker.lorem.words(5),
     excerpt: data.excerpt ?? faker.lorem.sentences(2),
     // if `null` is passed in, do not add a note to the list item
     note: data.note === undefined ? faker.lorem.sentences(2) : data.note,
-    imageUrl: data.imageUrl ?? faker.image.cats(),
+    imageUrl: data.imageUrl ?? faker.image.urlLoremFlickr({ category: 'cats' }),
```
jpetto, line 37: amazing new method name here 😆

jpetto (changes requested):
only real (small) blocker is typing of a couple functions return values. also a question on maybe making the query a bit more efficient.

thanks for doing all the faker updates!

src/shared/resolvers/fields/SavedItem.ts
katerinachinnappan: hehe yes, good point!

jpetto (approved):
this looks great! nice work!

PR #223, opened 2023-05-30
feat(save delete handler): beginnings of save delete handler

jpetto:
## Goal

beginnings of save delete handler

- console log for now to verify message format

## Tickets

- https://getpocket.atlassian.net/browse/OSL-520

## Implementation Decisions

just logging for now to validate message format/contents.

### DIFF
--- a/lambda/events/handlers/index.ts
+++ b/lambda/events/handlers/index.ts
@@ -1,8 +1,11 @@
 import { SQSRecord } from 'aws-lambda';
 import { accountDeleteHandler } from './accountDelete';
+import { saveDeleteHandler } from './saveDelete';
 
+// right hand value should map to the respective `detailType` in event bridge
 export enum Event {
   ACCOUNT_DELETION = 'account-deletion',
+  DELETE_ITEM = 'DELETE_ITEM',
 }
 
 // Mapping of detail-type (via event bridge message)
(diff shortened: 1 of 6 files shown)


lambda/events/handlers/index.ts
```diff
@@ -1,8 +1,11 @@
 import { SQSRecord } from 'aws-lambda';
 import { accountDeleteHandler } from './accountDelete';
+import { saveDeleteHandler } from './saveDelete';
 
+// right hand value should map to the respective `detailType` in event bridge
 export enum Event {
   ACCOUNT_DELETION = 'account-deletion',
+  DELETE_ITEM = 'DELETE_ITEM',
```
Herraj, line 8: Just making sure this is intentional since the `ACCOUNT_DELETION` key-value is different style?

Herraj (approved):

lambda/events/handlers/index.ts
jpetto: i'm _pretty_ sure this is accurate, as it's [defined here](https://github.com/Pocket/pocket-event-bridge/blob/main/.aws/src/event-rules/list-api-events/eventConfig.ts) in pocket event bridge. but yeah, good question, as the formats are _very_ different.

PR #224, opened 2023-05-30
Infra 1299/reusable infra

cmharlow:
Testing out reusable tf module setup for synthetic checks. Should not create anything new, but smoothly replace (after some manual tf imports) our singleton tf infra setup.


cmharlow:
the prod tf plan, with resources imported so it just shows meaningful changes (not recreating resources due to switch to modules): https://us-east-1.console.aws.amazon.com/codesuite/codebuild/996905175585/projects/ShareableListsApi-PRs-Prod/build/ShareableListsApi-PRs-Prod%3Afa46f432-fd63-44d6-8791-c3500e1d4225?region=us-east-1

.aws/src/main.ts
```diff
@@ -100,31 +100,32 @@ class ShareableListsAPI extends TerraformStack {
...
-      },
-      config.environment === 'Prod'
-        ? shareableListPagerduty.snsCriticalAlarmTopic.arn
-        : ''
-    );
+    new PocketAwsSyntheticChecks(this, 'synthetics', {
+      alarmTopicArn:
+        config.environment === 'Prod'
+          ? shareableListPagerduty.snsCriticalAlarmTopic.arn
+          : '', // this should be improved, empty string recreates updates constantly as is in cdktf
```
nina-py, line 107: Not a blocker, but wondering if we should have something set up on Dev and default to that instead of an empty string

nina-py (approved):

.aws/src/main.ts
cmharlow: I don't think Dev should be setup to page anything myself, even if its a page channel we ignore - I'd like to make all paging channels meaningful if we can.

cmharlow: (or even communication channels - like this could go to email for example, but we already get so much email we ignore, it seems not useful to point dev alerting to it)

cmharlow: but that's just my opinion.

PR #238, opened 2023-06-09
OSL-519: adding deleteShareableListItems express endpoint

katerinachinnappan:
## Goal

* Adding `deleteShareableListItems` express endpoint which deletes list items for a given user by `url` (coming from a `SavedItem`)

This endpoint will be called from a lambda function.

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-519](https://getpocket.atlassian.net/browse/OSL-519)

### DIFF
--- a/src/public/routes/deleteShareableListItems.ts
+++ b/src/public/routes/deleteShareableListItems.ts
@@ -26,6 +26,7 @@ const deleteShareableListItemDataSchema: Schema = {
 /**
  * This method batch deletes shareable list items by url for a user
  * @param userId
+ * @param url
  */
 export async function batchDeleteAllListItemsForUser(
   userId: number | bigint,
@@ -41,6 +42,7 @@ export async function batchDeleteAllListItemsForUser(
 /**
  * This method deletes all shareable list item data for a user by url
  * @param userId
+ * @param url
  */
 async function deleteShareableListItemUserData(
   userId: number | bigint,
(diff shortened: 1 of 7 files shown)


src/public/routes/deleteShareableListItems.ts
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * This method batch deletes shareable list items by url for a user
+ * @param userId
+ * @param url
+ */
+export async function batchDeleteAllListItemsForUser(
+  userId: number | bigint,
+  url: string
+): Promise<number> {
+  const batchResult = await db.listItem.deleteMany({
```
efixler, line 35: There's a max limit for the number of items that a user can have right?

efixler:
lgtm, tests look good.

We should be careful about multi-row updates/deletes without LIMITs as these can choke db. I think that right now there's a cap on list length that probably makes this ok, but it should like be revised to batch in the future.

src/public/routes/deleteShareableListItems.ts
katerinachinnappan: There have been conversations around this, but no concrete # yet. Not sure what's the decision on this one.

jpetto: we're waiting on product for a number here, but i am _pretty_ sure that we are going to impose a limit due to moderation constraints.

jpetto: but more to the point of this PR - there's no limit on the number of lists a user can have, so the url being deleted _could_ be in thousands of lists. good call on batching as a follow-up to make this process more robust/scalable.

jpetto (approved):
lgtm!

PR #240, opened 2023-06-13
feat(list): bump list updatedAt value on all related operations

jpetto:
## Goal

bump list `updatedAt` value on all related operations.

## Tickets

- https://getpocket.atlassian.net/browse/OSL-532

### DIFF
--- a/src/database/mutations/ShareableListItem.ts
+++ b/src/database/mutations/ShareableListItem.ts
@@ -75,9 +75,25 @@ export async function createShareableListItem(
     listId: list.id,
   };
 
-  const listItem = await db.listItem.create({
-    data: input,
-    select: shareableListItemSelectFields,
+  let listItem;
+
+  // execute the create and parent list updatedAt update in a single
+  // transaction
+  await db.$transaction(async (db) => {
+    listItem = await db.listItem.create({
+      data: input,
+      select: shareableListItemSelectFields,
+    });
+
+    // update the `updatedAt` field on the parent list
+    await db.list.update({
+      data: {
+        updatedAt: new Date().toISOString(),
+      },
+      where: {
+        externalId: list.externalId,
+      },
+    });
   });
 
   //send event bridge event for shareable-list-item-created event type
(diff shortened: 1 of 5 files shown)


src/database/mutations/ShareableListItem.ts
```diff
@@ -197,8 +226,26 @@ export async function updateShareableListItems(
...
+    for (const externalId of updatedShareableListExternalIds) {
+      await db.list.update({
+        data: {
+          updatedAt: new Date().toISOString(),
+        },
+        where: {
+          externalId,
+        },
+      });
+    }
```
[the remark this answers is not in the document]
katerinachinnappan: I think this route is ok, I rather update each list once than multiple times.

src/database/mutations/ShareableListItem.ts
```diff
@@ -253,21 +300,34 @@ export async function deleteShareableListItem(
...
-    })
-    .catch((error) => {
-      if (error.code === PRISMA_RECORD_NOT_FOUND) {
-        throw new NotFoundError(`List Item ${externalId} cannot be found.`);
-      } else {
-        // some unexpected DB error
-        throw error;
-      }
+  // the delete and parent list update should happen in a single transaction
+  await db.$transaction(async (db) => {
```
katerinachinnappan, line 304: Thnx for using the transaction API here

src/database/mutations/ShareableListItem.ts
```diff
@@ -125,11 +141,19 @@ export async function updateShareableListItem(
     delete data.sortOrder;
   }
 
+  // update the list item as well as the updatedAt value on the parent list
```
katerinachinnappan, line 144: Should we also wrap this block of update logic into a transaction?

katerinachinnappan (changes requested):

src/database/mutations/ShareableListItem.ts
jpetto: i _think_ since this is in a single prisma call, we don't need to use the transaction API 😅

katerinachinnappan: Oh what I meant was to wrap the entire block of code in a transaction. Updating list item, pushing listIds to an array and updating the list as it's all part of the update list item logic.

katerinachinnappan (approved):

src/database/mutations/ShareableListItem.ts
jpetto: oh no i'm confused. are you talking about the next mutation (`updateShareableListItems`) beginning on line 181? if so, that one _is_ in a transaction (but kind of hard to tell bc it's a lot of code). if you _are_ referencing this code (on line 144), then i might need more of an explanation 😄. happy to jump on a zoom!

katerinachinnappan: @jpetto nvm, the code did look confusing (was checking on the mobile UI while eating). Just checked the raw file separately & looks good , you're making one update call so doesn't need to be in a transaction. My bad, please merge!

PR #241, opened 2023-06-14
feat(ops work): some ops cleanup  - logger, cleanup, aggregation

cmharlow:
## Goal

This PR does some basic operational setup:

1. updates tf modules so the ALB Dashboard isn't overridden by terraform, so it can be used more generically as a reliability dashboard
2. removes the xray / opentelemetry stuff since it was only partially setup & unusable in current state
3. adds serverLogger setup using pocket-tools/ts-logger with some logging aggregation cleanup in cloudwatch for the ECS App

Notes:
* this doesn't add [context-aware child logger off the serverLogger object](https://github.com/winstonjs/winston#creating-child-loggers), but with request-specific details (e.g. userid, traceid, requestid pulled from HTTP request headers).  This would be a next step, but requires also updating all functions we would want request-specific logs from to be able to accept this ctxLogger object.
* this doesn't add the logger to the lambda - we could, but the lambda code size then becomes too big to ship.


src/database/client.ts
```diff
@@ -1,15 +1,46 @@
...
+  prisma.$on('warn', (e) => {
+    serverLogger.warn(e);
+  });
+
+  prisma.$on('info', (e) => {
+    serverLogger.info(e);
+  });
+
+  prisma.$on('query', (e) => {
+    serverLogger.debug(e);
```
jpetto, line 43: would be great to add a "prisma" prefix/source of some sort for slightly easier log filtering 😄

jpetto (changes requested):
one ask on log message prefixing, but otherwise awesome!

PR #250, opened 2023-07-10
OSL-576: send list item metadata to Snowplow

katerinachinnappan:
## Goal

This is part of the foundational refactor rollback. Adding back the list item metadata to Snowplow type and sending to event bridge.

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-576](https://getpocket.atlassian.net/browse/OSL-576)

### DIFF
--- a/src/snowplow/events.ts
+++ b/src/snowplow/events.ts
@@ -5,10 +5,7 @@ import {
 import * as Sentry from '@sentry/node';
 import config from '../config/';
 import { eventBridgeClient } from '../aws/eventBridgeClient';
-import {
-  ShareableListComplete,
-  ShareableListItemTemp,
-} from '../database/types';
+import { ShareableListComplete, ShareableListItem } from '../database/types';
 import {
   EventBridgeEventType,
   EventBridgeEventOptions,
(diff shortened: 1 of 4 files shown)


src/snowplow/events.ts
```diff
@@ -72,14 +69,25 @@ function transformAPIShareableListToSnowplowShareableList(
...
-  shareableListItem: ShareableListItemTemp,
+  shareableListItem: ShareableListItem,
   externalId: string,
   listExternalId: string
 ): SnowplowShareableListItem {
   return {
     shareable_list_item_external_id: externalId,
     shareable_list_external_id: listExternalId,
     given_url: shareableListItem.url,
+    title: shareableListItem.title ? shareableListItem.title : undefined,
```
jpetto, line 80: no need to refactor, but the logical OR operator could work here for slightly less typing. e.g.

`title: shareableListItem.title || undefined,`

(or if we wanted to pass through empty strings/`0` values, the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing).)

jpetto (approved):

src/snowplow/events.ts
katerinachinnappan: yep good point, lets refactor this in the next PR, I will make a note!

PR #257, opened 2023-07-17
OSL-577: block ability to make lists public

katerinachinnappan:
## Goal

Blocking the ability to make a list `PUBLIC`. Current `PUBLIC` lists can stay `PUBLIC`, but if switched to `PRIVATE`, can no longer switch back to `PUBLIC`. Commented out relevant tests in case we ever turn back the functionality. 

## Tickets

- [https://getpocket.atlassian.net/browse/OSL-577](https://getpocket.atlassian.net/browse/OSL-577)

### DIFF
--- a/.circleci/scripts/setup_aws.sh
+++ b/.circleci/scripts/setup_aws.sh
@@ -2,7 +2,9 @@
 set -e
 
 sudo apt-get update && sudo apt-get install -y python3-pip
-pip3 install awscli-local awscli
+# using --no-build-isolation flag for now, remove when issue is fixed (AWS CLI)
+# https://github.com/aws/aws-cli/issues/8036
+pip3 install awscli-local awscli --no-build-isolation
 
 for Script in .docker/localstack/*.sh ; do
     bash "$Script"
--- a/src/database/mutations/ShareableList.ts
+++ b/src/database/mutations/ShareableList.ts
@@ -4,6 +4,7 @@ import {
   UserInputError,
 } from '@pocket-tools/apollo-utils';
 import { Visibility, ModerationStatus, PrismaClient } from '@prisma/client';
+// import slugify from 'slugify';
 import {
   CreateShareableListInput,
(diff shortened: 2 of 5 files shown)


.circleci/scripts/setup_aws.sh
```diff
@@ -2,7 +2,9 @@
 set -e
 
 sudo apt-get update && sudo apt-get install -y python3-pip
-pip3 install awscli-local awscli
+# using --no-build-isolation flag for now, remove when issue is fixed (AWS CLI)
+# https://github.com/aws/aws-cli/issues/8036
```
jpetto, line 6: 🙏

src/database/mutations/ShareableList.ts
```diff
@@ -115,17 +113,22 @@ export async function updateShareableList(
     throw new NotFoundError(`A list by that ID could not be found`);
   }
 
+  // we don't allow lists to be public anymore. block updates that try to make a list public
+  if (data.status === Visibility.PUBLIC) {
+    throw new ForbiddenError(ACCESS_DENIED_ERROR);
+  }
```
jpetto, line 119: this might need to be a check against the current value and the proposed value. if the list is currently `PUBLIC` and is edited with that same status, the update should succeed.

src/database/mutations/ShareableList.ts
```diff
@@ -21,8 +20,7 @@ import {
   ACCESS_DENIED_ERROR,
   PRISMA_RECORD_NOT_FOUND,
 } from '../../shared/constants';
-import { getShareableList, isPilotUser } from '../queries';
```
jpetto, line 24: for longer-term maintenance, would be rad to comment on all the imports here that are no longer used anywhere - like the slugify stuff, and i think `isPilotUser`? could comment them out with a link to this ticket.

jpetto (changes requested):
one functionality request and a little cleanup task 😄

